PR #56, opened 2014-09-03
Feature/call makefile from anywhere

syclik:
A model can now be built from any directory by calling make with "-f <CMDSTAN>/makefile".

All of the CmdStan targets can be run from any directory with the same "-f" flag to make.

Needs additional user documentation.


syclik:
I'm closing this because we're not going to get around the limitation of spaces in paths using make.

PR #117, opened 2015-06-24
Feature/advi command

dustinvtran:
#### Summary:
Adds arguments to `command.hpp` for running variational inference. To be merged when `feature/bbvb` in the main repo is.

#### Intended Effect:
Can use variational inference in cmdstan.

#### How to Verify:
Run models by building them, then running them, e.g. `make examples/bernoulli/bernoulli && cd examples/bernoulli && ./bernoulli variational data file=bernoulli.data.R`.

#### Side Effects:
N/A

#### Documentation:
Adds to the cmdstan manual with all the arguments users can specify for running variational inference.

#### Reviewer Suggestions:
N/A


syclik:
This looks good. Let me update to `develop` and see if it works.

syclik:
Before this gets merged, I'd like to see a test that runs ADVI. You can model it off of something in src/test/interface/*.cpp.

(We don't have enough tests in there that test end-to-end behavior now, but since you're adding something, we can start here. If you want some help, we can discuss Thursday.)

syclik:
@dustinvtran and @akucukelbir, this failure is real. Not sure what's going on yet.

syclik:
To run the test that fails locally:
```
> make test/interface/services/argument_configuration_test
> test/interface/services/argument_configuration_test
```

Once that test is built, you can debug a little easier by doing this:
```
> cd test
>  ../src/test/test-models/test_model method=variational algorithm=fullrank iter=10000 grad_samples=1 elbo_samples=100 eta_adagrad=0.10000000000000001 tol_rel_obj=0.01 eval_elbo=100 output_samples=1000
```

Depending on the run, I get:
```
...
  iter       ELBO   delta_ELBO_mean   delta_ELBO_med   notes 

Exception: stan::variational::normal_fullrank::set_L_chol: Dimension of mean vector (2) and Dimension of input matrix (4) must match in size
Diagnostic information: 
Dynamic exception type: std::invalid_argument
std::exception::what: stan::variational::normal_fullrank::set_L_chol: Dimension of mean vector (2) and Dimension of input matrix (4) must match in size
```

and 

```
...
  iter       ELBO   delta_ELBO_mean   delta_ELBO_med   notes 
test_model(61234,0x7fff741f0300) malloc: *** error for object 0x7fc41b5005c8: incorrect checksum for freed object - object was probably modified after being freed.
*** set a breakpoint in malloc_error_break to debug
Abort trap: 6
```

This is for fullrank. Mean field seems to work just fine.

akucukelbir:
@dustinvtran somewhere there's a `dimension_` that's getting squared. (i just checked for size 3 and it complains about size 9.) either that, or the matrix is getting flattened into a vector. (this is more likely.)

the function that fails is `set_L_chol` which is only called in https://github.com/stan-dev/stan/blob/develop/src/stan/variational/families/normal_fullrank.hpp#L287 the parameter `L_grad` is updated just above that line in 3 places. the code looks the same as it did, back when it worked.

my current guess is that perhaps there's a bug in the overloading of the arithmetic operators that you implemented when templating out the variational family? i'm not sure. but somehow `L_grad` no longer has size `dimension_` but rather `dimension_ * dimension_`.

dustinvtran:
I spotted the bug, which turned out to be a trivial problem as a result of the type that `L_grad` was set as. [Pull request](https://github.com/stan-dev/stan/pull/1532) submitted in the stan repo.

bob-carpenter:
I'll try to track that down today.  Enjoy the conference!

PR #125, opened 2015-07-12
Feature/issue advi lp

dustinvtran:
#### Summary:
Renames `lp` to `lp__` in output of variational inference.

#### Intended Effect:
N/A

#### How to Verify:
Run models by building them, then running them, e.g. `make examples/bernoulli/bernoulli && cd examples/bernoulli && ./bernoulli variational data file=bernoulli.data.R`.

#### Side Effects:
N/A

#### Documentation:
Changed variational inference documentation to reflect that `lp__` is actually calculated now. See [stan pull request](https://github.com/stan-dev/stan/pull/1539)

#### Reviewer Suggestions:
N/A

### DIFF
--- a/src/test/interface/variational_output_test.cpp
+++ b/src/test/interface/variational_output_test.cpp
@@ -1,0 +1,74 @@
+#include <gtest/gtest.h>
+#include <test/utility.hpp>
+#include <stan/mcmc/chains.hpp>
+#include <fstream>
+
+using cmdstan::test::convert_model_path;
+using cmdstan::test::run_command;
+using cmdstan::test::run_command_output;
+
+class CmdStan : public testing::Test {
+ public:
+  void SetUp() {
+    std::vector<std::string> model_path;
+    model_path.push_back("src");
+    model_path.push_back("test");
+    model_path.push_back("test-models");
+    model_path.push_back("variational_output");
+
+    output_file = "test/output.csv";
+
+    base_command = convert_model_path(model_path)
+      + " output file=" + output_file;
+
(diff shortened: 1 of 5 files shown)


syclik:
Jenkins, ok to test.

dustinvtran:
Just realized that the checks never passed for this pull request; I added the unit test to the `make/tests` file. Please re-review this and merge :dancer:

src/test/interface/variational_output_test.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+
+TEST_F(CmdStan, variational_default) {
+  run_command_output out = run_command(base_command + " variational");
+
+  ASSERT_EQ(0, out.err_code);
+
+  stan::mcmc::chains<> chains = parse_output_file();
+  ASSERT_EQ(1, chains.num_chains());
+  ASSERT_EQ(1001, chains.num_samples());
```
syclik: is this all we want to test?

syclik:
Sorry -- I missed this from 3 weeks ago. Is that all you want to test? Just make sure it works? Or did you want to make sure it's non-zero (to prevent a bug of someone accidentally changing the underlying implementation)? Or something else?

dustinvtran:
I think it makes sense to only check if the output exists. `lp__` is temporary and only exists in the output because it needs to be compatible for `write_iteration_csv()`. Ideally, we plan on storing other things/replacing things, and we'll need to make a unit test at that time after we've all decided on the output interface for ADVI.

PR #166, opened 2015-10-13
Change argument of eta_adagrad to eta

dustinvtran:
#### Summary:
See https://github.com/stan-dev/stan/pull/1638.


dustinvtran:
Just to be clear: the checks are failing because it requires stan-dev/stan#1638 to be merged.

PR #222, opened 2015-11-19
Puts argument output back into the output file

syclik:
#### Summary:

Puts argument output back into the output file. This was accidentally taken out between 2.8.0 and now.

#### Intended Effect:

Makes CmdStan behave as it did in the past.

#### How to Verify:

Run the tests. Verify output.

The test should catch some of the issues in the future.

#### Side Effects:

None.

#### Documentation:

None.

#### Reviewer Suggestions: 

@akucukelbir, can you verify?

### DIFF
--- a/src/cmdstan/command.hpp
+++ b/src/cmdstan/command.hpp
@@ -245,8 +245,8 @@ namespace stan {
       if (output_stream) {
         io::write_stan(output_stream, "#");
         io::write_model(output_stream, model.model_name(), "#");
-        parser.print(info);
-        info();
+        interface_callbacks::writer::stream_writer sample_writer(*output_stream, "# ");
+        parser.print(sample_writer);
       }
 
       if (diagnostic_stream) {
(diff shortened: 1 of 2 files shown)


akucukelbir:
@syclik verified. merge at will.

src/cmdstan/command.hpp
```diff
@@ -242,11 +242,14 @@ namespace stan {
...
+      parser.print(info);
+      info();
+      
       if (output_stream) {
         io::write_stan(output_stream, "#");
         io::write_model(output_stream, model.model_name(), "#");
-        parser.print(info);
-        info();
+        interface_callbacks::writer::stream_writer sample_writer(*output_stream, "#");
+        parser.print(sample_writer);
```
bob-carpenter: I take it this is intentional --- only write to `sample_writer` if `output_stream` isn't `NULL`?

bob-carpenter:
Assuming the conditioning was intentional, looks OK to me.

syclik:
Yes, that was intentional. Once the full refactor is done, we'll be able to
collect all the writers to the top of the control loop, instantiate once,
and move forward.

PR #265, opened 2015-12-10
Print currently printing with '#'

syclik:
#### Summary:

Print statements within a Stan program in the model block prints with a leading `# `. This removes it.

#### Intended Effect:

Remove the `# ` from printing.

#### How to Verify:

Run the new test.

#### Side Effects:

None. This matches behavior to the last tagged version, v2.9.0.

#### Documentation:

None.

#### Reviewer Suggestions: 

None.


bob-carpenter:
Looks OK to me.

PR #512, opened 2016-11-25
Feature/allow undefined

bgoodri:
#### Submisison Checklist

- [X] Run tests: `./runCmdStanTests.py src/test`

#### Summary:

Pass the environmental variable `ALLOW_UNDEFINED` to `stanc`. 
Also, pass `-include path/to/USER_HEADER.hpp` to the compiler.

#### Intended Effect:

If the user has `ALLOW_UNDEFINED=--allow_undefined` in `make/local` then models will parse with functions that are declared but not defined.
If USER_HEADER.hpp is written correctly, then when the Stan program executes, it will use the functions defined therein.

#### How to Verify:

Put `ALLOW_UNDEFINED=--allow_undefined` in `make/local` and do `./runCmdStanTests.py src/test` but that doesn't actually do much because there are no test files in cmdstan that have undefined functions (there is a parser test in the Stan library already). 

#### Side Effects:

Linker errors. At the moment, models with functions that are declared but not defined will parse and compile but not link. The user either needs to edit a USER_HEADER.hpp file or specify `LDLIBS = -Lmy_lib_dir -lmy_lib` in `make/local` to link to a shared object called `my_lib.o`. Then it should compile. 

#### Documentation:

In section 4.4 of cmdstan-guide.pdf

#### Reviewer Suggestions: 

@syclik 

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company): Trustees of Columbia University

### DIFF
--- a/make/models
+++ b/make/models
@@ -8,7 +8,8 @@ CMDSTAN_MAIN := src/cmdstan/main.cpp
 $(patsubst %.stan,%,$(wildcard $(addsuffix .stan,$(MAKECMDGOALS)))) : %$(EXE) : %.hpp %.stan bin/stanc$(EXE) bin/stansummary$(EXE) $(LIBCVODES)
 	@echo ''
 	@echo '--- Linking C++ model ---'
-	$(LINK.c) -O$O $(OUTPUT_OPTION) $(CMDSTAN_MAIN) -include $< $(LIBCVODES) $(LDLIBS)
+	@test -f $(dir $<)USER_HEADER.hpp || touch $(dir $<)USER_HEADER.hpp
+	$(LINK.c) -O$O $(OUTPUT_OPTION) $(CMDSTAN_MAIN) -include $< -include $(dir $<)USER_HEADER.hpp $(LIBCVODES) $(LDLIBS)
 
 .PRECIOUS: %.hpp
 %.hpp : %.stan $(MODEL_HEADER) bin/stanc$(EXE)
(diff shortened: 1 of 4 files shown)


syclik:
Sorry -- I didn't realize this PR existed. I'll look over it tonight.

syclik:
@bgoodri, we need to fix the behavior. This pull request drops an extra file into the directory. We shouldn't do that to users. I don't care that the name of the file is hardcoded, but we really shouldn't be just dropping files that the user didn't create there just because we need it as a default.

syclik:
I'll think of a way to make that work in general. I don't know if you know of any better makefile logic.

bgoodri:
That is how I did it originally, but then it does not compile until the
user edits the C++ file.

syclik:
What do you mean? I'm ok with reading from a file even if the name is
hard-coded. I'm not ok with creating a file when it doesn't exist. We
should just switch on a flag, I think.

But... I'm still uncertain what you mean. Maybe I'm missing something.

bgoodri:
We could pass a flag like USER_HEADER=/path/to/file to make, but then we
have to make changes to JuliaStan etc. to know when to do that.

syclik:
We could do it in the same way you have it implemented. Perhaps something
like: ALLOW_UNDEFINED=TRUE. Or something similar.

That could either go on the command line or make/local.

If that's true, then we could call the compiler with the right flags, make
sure to rebuild if the file is present, etc.

I don't like that the file is forced to be there and if it isn't present, a
new file is dropped there.

syclik:
@bgoodri, can you take a look at this pull request now and see if this does what you want it to do? I changed it to have two extra make flags.

bgoodri:
Did you push?

syclik:
Thanks for the reminder. Just did.

syclik:
(public internet cut out when I tried to push and I didn't notice... my bad)

syclik:
@bgoodri, let me know if this is good. This is the last PR I'm waiting on for v2.13.1.

bgoodri:
It looks fine. I just changed the doc a little.

Issue #540, opened 2017-04-03
No automated diagnostics


betanalpha:
#### Description:

CmdStan currently doesn't have any facilities to diagnose pathological MCMC output.

#### Current Version:
v2.14.0

PR #541, opened 2017-04-03
Prototypical diagnostic facility

betanalpha:
#### Submisison Checklist

#### Summary:

Introduces a prototypical diagnostic facility.  The exact text and thresholds can be modified per user feedback.  Addresses #540.

#### Intended Effect:

Provides a compliment to `bin/stansummary` that analyzes the output instead of simply displaying it.

#### How to Verify:

`./runCmdStanTests.py src/test/interface/diagnose_test.cpp`

#### Side Effects:

None.

#### Documentation:

None added -- we should iterate a few times before exposing this to users via the manual.

#### Reviewer Suggestions: 

@syclik 

#### Copyright and Licensing


make/command
```diff
@@ -12,3 +12,8 @@ bin/print$(EXE) : bin/cmdstan/print.o
 bin/stansummary$(EXE) : bin/cmdstan/stansummary.o
 	@mkdir -p $(dir $@)
 	$(LINK.c) -O$(O_STANC) $(OUTPUT_OPTION) $< $(LDLIBS)
+
+.PRECIOUS: bin/diagnose$(EXE)
```
syclik: You'll want to add the `bin/diagnose` target to the `build` target defined in `/makefile`

You'll also want to add it to the model build step in: `make/models`.

We might be able to simplify the makefiles somehow, but that's for a different PR.

src/cmdstan/diagnose.cpp
```diff
@@ -0,0 +1,189 @@
...
+  std::vector<std::string> filenames;
+
+  for (int i = 1; i < argc; i++) {
+    ifstream.open(argv[i]);
+
+    if (ifstream.good()) {
+      filenames.push_back(argv[i]);
+      ifstream.close();
+    } else {
+      std::cout << "File " << argv[i] << " not found" << std::endl;
```
syclik: There are a few other ways this can fail. Did we want to get any fancier with the error reporting here?

src/cmdstan/diagnose.cpp
```diff
@@ -0,0 +1,189 @@
...
+  if (argc == 1) {
+    diagnose_usage();
+    return 0;
+  }
+
+  // Parse any arguments specifying filenames
+  std::ifstream ifstream;
+  std::vector<std::string> filenames;
+
+  for (int i = 1; i < argc; i++) {
```
syclik: pedantic: `++i`

src/cmdstan/diagnose.cpp
```diff
@@ -0,0 +1,189 @@
...
+  ifstream.open(filenames[0].c_str());
+
+  stan::io::stan_csv stan_csv
+    = stan::io::stan_csv_reader::parse(ifstream, &std::cout);
+  stan::mcmc::chains<> chains(stan_csv);
+
+  ifstream.close();
+
+  for (std::vector<std::string>::size_type chain = 1;
+       chain < filenames.size(); chain++) {
```
syclik: pedantic: `++chain`

src/cmdstan/diagnose.cpp
```diff
@@ -0,0 +1,189 @@
...
+    ifstream.close();
+  }
+
+  int num_samples = chains.num_samples();
+
+  std::vector<std::string> bad_n_eff_names;
+  std::vector<std::string> bad_rhat_names;
+
+  for (int i = 0; i < chains.num_params(); i++) {
+    int max_limit = 10;
```
syclik: It looks like this is related to max treedepth. Should this be controlled from the outside somehow?

src/cmdstan/diagnose.cpp
betanalpha: stan_csv_reader will throw if the files are ill-formatted.  It's the same error handling as in stansummary.

betanalpha:
Changes pushed.  Right now the only thing that could be pushed to Stan is the E-BFMI calculation, but even that would be awkward given that the chains object doesn't discriminate between its variables.  At the moment I think it's best to keep it in the interfaces.

max_limit should be the max_treedepth argument for the input chains, but that's not saved in the stan_csv_reader so it cannot be accessed programmatically.  I went with 10 as it's the CmdStan default.

bob-carpenter:
Cool!  Just waiting on @syclik's approval for the changes.

PR #582, opened 2017-11-02
Use precompiled headers to shave off ~15% compile time for models.

seantalts:
#### Submisison Checklist

#### Summary:
Change CmdStan to use precompiled headers. On my machine this cuts out about 4-5 seconds of compile time (once everything else is compiled and we're just compiling and linking the CPP model). I doubt this scales linearly with more complicated models but who knows! My bernoulli example takes about 24s without PCH and 19s with. @bgoodri might be worth porting to RStan?

#### Copyright and Licensing

### DIFF
--- a/make/models
+++ b/make/models
@@ -2,17 +2,21 @@
 # Models (to be passed through stanc)
 ##
 MODEL_HEADER := $(STAN)src/stan/model/model_header.hpp
+MODEL_PCH := $(MODEL_HEADER:.hpp=.hpp.gch)
 CMDSTAN_MAIN := src/cmdstan/main.cpp
 
+$(MODEL_PCH) : $(MODEL_HEADER)
+	@echo 'Compiling pre-compiled header'
+	$(COMPILE.cc) -O$O $(MODEL_HEADER) -o $@
 
 .PRECIOUS: %.hpp %.o
-$(patsubst %.stan,%,$(wildcard $(addsuffix .stan,$(MAKECMDGOALS)))) : %$(EXE) : %.hpp %.stan bin/stanc$(EXE) bin/stansummary$(EXE) bin/diagnose$(EXE) $(LIBCVODES)
+$(patsubst %.stan,%,$(wildcard $(addsuffix .stan,$(MAKECMDGOALS)))) : %$(EXE) : %.hpp %.stan bin/stanc$(EXE) bin/stansummary$(EXE) bin/diagnose$(EXE) $(LIBCVODES) $(MODEL_PCH)
 	@echo ''
 	@echo '--- Linking C++ model ---'
 ifneq (,$(findstring allow_undefined,$(STANCFLAGS)))
-	$(LINK.cc) -O$O $(OUTPUT_OPTION) $(CMDSTAN_MAIN) -include $< -include $(USER_HEADER) $(LIBCVODES)
+	$(LINK.cc) -include-pch $(MODEL_PCH) -O$O $(OUTPUT_OPTION) $(CMDSTAN_MAIN) -include $< -include $(USER_HEADER) $(LIBCVODES)
 else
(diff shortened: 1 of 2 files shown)


bgoodri:
The pch is too big for CRAN.

seantalts:
Can't we build it locally like with this patch? I don't fully understand how RStan works.

bgoodri:
CRAN makes a shared object, which last I checked would be about 250MB with
a PCH.

Something like this might work for a package like rstanarm, to create the
PCH first, compile all the models, and then throw the PCH away.

seantalts:
What does CRAN make a shared object for? I would imagine it would just make one for the stan compiler, but we still need the source code of the stan and math headers on disk in order to compile a Stan model, no? If we have those on disk we can always stop using them as .hpp files and use an amalgamated .hpp.gch precompiled file that we construct as part of the build process instead... I might still be missing something.

seantalts:
@bob-carpenter are you the best person to review this? Or maybe @mitzimorris?

syclik:
It's probably me. I'll take a look this weekend.

seantalts:
Maybe someone else has time to do it today? I have been putting a ton of PRs through to you and I think that's unfair given that this isn't your full time job anymore.

bob-carpenter:
I don't even know what language that's written in!

seantalts:
So an interesting question is - does this even need code review if Daniel was the only one who could read makefiles before?

sakrejda:
Strictly speaking I _can_ read that and it looks good.  I _can't_ read it well enough that I would trust that it would work as intended without verifying the actual compiler calls that are generated but I assume you did that.

seantalts:
Yep, and all the tests pass, it runs the same number of tests, that kind of thing. Want to approve it for me? We can always back it out later if someone else finds an issue with it. I don't think there's any "CmdStan czar" that needs to see every PR or anything.

seantalts:
Thanks!

syclik:
awesome! thanks, Krzysztof and Sean!

syclik:
Oh... I know it's merged now, but could you verify that it works on both Mac and Windows? I thought we had trouble with Windows g++ and precompiled headers.

seantalts:
Funny story - CmdStan does not seem to work on windows before this commit. I checked out the one before:

image

syclik:
You need to add .exe to the end of that command.

Yay, Windows!

sakrejda:
(╯°□°)╯︵ ┻━┻    <--- Windows

seantalts:
Bahaha. Do executables actually have to end in .exe on Windows??

On Fri, Nov 3, 2017 at 2:17 PM, Krzysztof Sakrejda <notifications@github.com

seantalts:
I have to say, that .exe trick works way better. Unfortunately, it still doesn't seem to totally work:
image

What else am I doing wrong? 😅

sakrejda:
Does the libsundials directory exist? Does the library exist?

seantalts:
The `lib` directory under the cvodes directory does not exist. Make does
not appear to have attempted to make them.

On Fri, Nov 3, 2017 at 4:11 PM, Krzysztof Sakrejda <notifications@github.com

seantalts:
Should I be using something other than git bash?

sakrejda:
Did you install g++ via Rtools?

syclik:
What does git bash have to do with anything? Are you using that instead of command.exe? If so, I don't know how that interacts with make. 

Can you try:
make clean-all (or use git to clean)
make build

Then build the executable?

seantalts:
I was trying both. `make build` seems to be a necessary precursor step,
even before this PR. Interestingly, the PR does not appear to influence
compilation on Windows at all?  I'm not sure how that works, but the PCH
stuff is just totally missing:
 [image: Inline image 1]

PR #589, opened 2017-12-08
Add suggestions to doc to look in make/local

seantalts:
And remove unused CC= declarations etc in the makefile that were
confusing people.

#### Submisison Checklist

#### Copyright and Licensing

### DIFF
--- a/makefile
+++ b/makefile
@@ -11,16 +11,13 @@ help:
 
 ##
 # Users should only need to set these three variables for use.
+# Please set them in a test file called make/local with no extension.
 # - CC: The compiler to use. Expecting g++ or clang++.
 # - O: Optimization level. Valid values are {0, 1, 2, 3}.
 # - AR: archiver (must specify for cross-compiling)
 # - OS: {mac, win, linux}. 
-##
-CC = g++
-O = 3
-O_STANC = 0
-AR = ar
-
+#
+#
 ##
 # Library locations
 ##
--- a/src/docs/cmdstan-guide/getting-started.tex
+++ b/src/docs/cmdstan-guide/getting-started.tex
@@ -910,7 +910,8 @@ and on Windows with
 %% \emph{Warning}: \ The unit tests can take 30+ minutes and consume 3+
 %% GB of memory with the default compiler, \code{g++}.  The distribution
 %% test and model tests can take even longer.  It is faster to run the
-%% Clang compiler (option \code{CC=clang++}), and to run in multiple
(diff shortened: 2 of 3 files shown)


makefile
```diff
@@ -11,16 +11,13 @@ help:
 
 ##
 # Users should only need to set these three variables for use.
+# Please set them in a test file called make/local with no extension.
 # - CC: The compiler to use. Expecting g++ or clang++.
```
syclik: Change `CC` to `CXX`.

src/docs/cmdstan-guide/getting-started.tex
```diff
@@ -910,7 +910,8 @@ \subsection{Configuring Command-Line Options}
 %% \emph{Warning}: \ The unit tests can take 30+ minutes and consume 3+
 %% GB of memory with the default compiler, \code{g++}.  The distribution
 %% test and model tests can take even longer.  It is faster to run the
-%% Clang compiler (option \code{CC=clang++}), and to run in multiple
+%% Clang compiler (option \code{CC=clang++} in make/local), and to run
```
syclik: change `CC` to `CXX`

syclik:
I don't think this actually fixes the issues with the builds. We should take care of that before releasing the next version.

seantalts:
What build issues are you referring to specifically?

And I think it’s supposed to be CC, actually, looking at how it’s used?

sakrejda:
Libc++flag contaminates builds that switch to GCC via CC variable, for
example

seantalts:
Oh, like people wanted to be able to compile more easily with stdlib instead of libc++?

sakrejda:
Not quite, the libc++ flag should be compiler-specific so if we're not
using clang++ it should not get pulled in at all.

seantalts:
There are major LTS versions of Ubuntu (used in Travis) with broken stdlib but working libc++, I think :( I think letting users set it is probably also important for that reason. Would probably make more sense to have the default library be the one that comes with the compiler you're using though. Let me open an issue for that.

sakrejda:
Yeah, I know, also versions of Ubuntu that can't successfully install clang++.... my point is we should make it straightforward to change these in ways that allow all working combos.  I tried to do this but I ended up making changes across math/stan/cmdstan makefiles and never backed out what the actual commits would need to be.

seantalts:
Gotcha. Thanks for giving it a shot! I think I can do a simple patch and await more ambitious refactorings later.

seantalts:
https://github.com/stan-dev/cmdstan/issues/590

Issue #590, opened 2017-12-11
Need better configuration for C++ standard library options.


seantalts:
#### Summary:
We'd like to let users easily compile with either gcc's stdlib or llvm's libc++ regardless of which compiler they're using (especially because these libraries are often broken in Ubuntu releases). We should probably also make the default standard library be whichever one comes with your compiler.

#### Current Version:
v2.17.0

seantalts:
@sakrejda @syclik Seems like with `make/local` set to `CC=g++` things work appropriately and it implicitly uses gcc's standard library? e.g.
```
g++ -Wall -I . -isystem stan/lib/stan_math/lib/eigen_3.3.3 -isystem stan/lib/stan_math/lib/boost_1.64.0 -isystem stan/lib/stan_math/lib/cvodes_2.9.0/include -std=c++1y -DBOOST_RESULT_OF_USE_TR1 -DBOOST_NO_DECLTYPE -DBOOST_DISABLE_ASSERTS -DBOOST_PHOENIX_NO_VARIADIC_EXPRESSION -Wno-unused-function -Wno-uninitialized -I src -I stan/src -isystem stan/lib/stan_math/ -DEIGEN_NO_DEBUG -DFUSION_MAX_VECTOR_SIZE=12 -DNO_FPRINTF_OUTPUT -pipe  -c -O0 -o bin/stan/lang/grammars/statement_2_grammar_inst.o stan/src/stan/lang/grammars/statement_2_grammar_inst.cpp
```

sakrejda:
Is this the develop branch? I'll check when I get home how it breaks for me (I had to add something in make/local but also edit the makefiles last time I checked...

seantalts:
Yep in develop. Thanks!

sakrejda:
@seantalts if I set CXX=g++ in make/local for cmdstan develop I get:

```
krzysztof@hagrid:~/packages/cmdstan-develop$ make examples/bernoulli/bernoulli
[...]

Issue #588, opened 2017-12-08
cmdstan 2.17.0 unable to compile the program after 'make build' in windows 10


cmelog:
#### Summary:
I had some issues but finally were able to build the cmdstan-2.17.0 on windows 10. Now I am trying to run the first example, bernoulli but without any success

#### Description:
After a while I was able to get:

--- CmdStan v2.17.0 built ---

But when I am trying to run the bernoulli example I got the following message:

C:\cmdstan-2.17.0>make examples/bernoulli/bernoulli.exe
uname: not found
uname: not found
make: Nothing to be done for `examples/bernoulli/bernoulli.exe'.

#### Reproducible Steps:
I modified the extension of the file 'bernoulli.stan' on the examples folder to 'bernoulli.exe' according to the guide on page 10. 

As I mentioned before I am trying to run the bernoulli example but I got:

make: Nothing to be done for `examples/bernoulli/bernoulli.exe'.

#### Current Output:

make: Nothing to be done for `examples/bernoulli/bernoulli.exe'.

#### Expected Output:
I am supposed to generate a C++ code bernoulli.hpp and the compiled executable. 

#### Additional Information:
Provide any additional information here.

#### Current Version:
v2.17.0

syclik:
@cmelog, I now understand your confusion.

1. change the name of the file from `bernoulli.exe` to `bernoulli.stan`
2. then type `make examples/bernoulli/bernouli.exe`

Please let me know if that works.

cmelog:
Hi, I changed the file to bernoulli.stan and got:

C:\cmdstan-2.17.0>make examples/bernoulli/bernouli.exe
uname: not found
uname: not found
make: *** No rule to make target `examples/bernoulli/bernouli.exe'.  Stop.

I am now running Stan from Mathlab and is working, which was my original goal. I am not sure if I should also be able to run on cmdstan

Ethiral:
Hello all,
I'm facing the same issue No rule to make target `C:/Users/EVerma/cmdstan/examples/bernoulli/bernoulli.exe'
How did you resolve this?
Thanks

cmelog:
Hi this is what I did:

1. Installed the newest version of Rtools.
2. On the path I included: C:\Rtools\mingw_64\bin to select the adequate version of the Rtools.
3. I created a local file (no extension)  in the C:\cmdstan-2.17.1\make with the following lines:
CC=g++

I hope this helps.

Ethiral:
[...]

Issue #586, opened 2017-11-27
Error installing cmdstan in linux: it’s trying to use clang 


syclik:
Thanks, bmajoros (on Discourse)! See original thread: http://discourse.mc-stan.org/t/error-installing-cmdstan-in-linux-its-trying-to-use-clang/2620

#### Summary:
Even though the makefile appears to have g++ set as the compiler, it’s trying to call clang++, and that compiler is not installed.

#### Description:
CmdStan is not respecting the make options.
(I've seen this on another install.)

#### Reproducible Steps:
From a linux install without `clang++` installed, setting `CC=g++` or `CXX=g++` fails to run the `make build` target.

#### Current Output:
```
make: clang++: Command not found
```

#### Expected Output:
Everything should build.

#### Additional Information:
We changed the makefiles in Math and Stan. I think this issue stems from not handling the includes properly (but it could be anything really since it's make).

#### Current Version:
v2.17.0

seantalts:
I think this is a duplicate of https://github.com/stan-dev/cmdstan/issues/566

PR #598, opened 2017-12-13
Fixing build issues

syclik:
#### Submisison Checklist

#### Summary:

Fixes a lot of build issues.

This allows users in Windows, Linux, and Mac to run CmdStan without additional configuration. The defaults allow models to be built without additional configuration.

#### Intended Effect:

Fixes build issues that users were reporting:
- #586 
- #590 
- #588 

#### How to Verify:

Build from:
- Mac, Windows, and Linux
- from a fresh clone without a `make/local` or `~/.config/cmdstan/make.local` file present

And build the bernoulli example.

Please verify that this works on your local machine or whatever other machine you have access to.

#### Side Effects:

These are the known side-effects in this pull request. These shouldn't affect users. The second one may affect developers:
- Mac: on develop, we have included in the `CXXFLAGS` `-stdlib=libc++`. In this branch, this is omitted. 
    Without it, it builds fine on my machine. I removed the default compiler from being set to `clang++`. Instead, we use the system default, which usually defaults to `c++`. The math library's `detect_cc` searches for the string `clang++` and uses that to set options.
- Linux: tests don't build with the default values. In order to build them, you need to include `GTEST_CXXFLAGS += -DGTEST_HAS_PTHREAD=0` in `make/local` or `~/.config/cmdstan/make.local`.
    This is also happening because `CXX`'s default value is `c++`. The `os_linux` file uses the value there (indirectly) to add to the `GTEST_CXXFLAGS`. We should probably just add that directly for all of Linux, but the fix should be in the Math library. This has been documented.

#### Documentation:

- Added subsection mentioning how to build tests in linux. 
- Minor updates to change `CC` to `CXX` (this may conflict with the other PR, but we can work through that easily).

#### Copyright and Licensing

Generable


seantalts:
Hey, I think there's a lot of confusion going on about what build issues actually exist - check out the post here: https://github.com/stan-dev/cmdstan/issues/590#issuecomment-351215438 

Maybe we can hop on a hangout or something? I don't actually think there are any build issues at this point (now that we finally merged that PR to fix the pch thing for g++ people), assuming we doc correctly how to set CC in make/local.

Also, I think it's probably a good idea to switch to `g++` as the default compiler everywhere (I'm setting up a PR in math repo) because that's aliased on macs to clang++, so it will work out of the box for many more people.

syclik:
Sure, we can jump on a hangout.

Re: build issues. We have a few with v2.17.0. What I expect is in mac,
linux, and windows, with a normal-ish setup, it would run out of the box.
Currently, that doesn't happen for linux and windows. That's what I was
trying to fix here.

src/docs/cmdstan-guide/installation.tex
```diff
@@ -454,6 +454,20 @@ \subsection{Downloading and Unpacking \CmdStan Source}
...
 \end{quote}
 
+\subsection{Running \CmdStan Tests}
+
+Google test uses pthread by default. If the library is not available on your
+system, it's still easy to run the tests. Add this to your \code{make/local}
+file:
+\begin{quote}
+\begin{Verbatim}[fontshape=sl,fontsize=\small]
+GTEST_CXXFLAGS += -DGTEST_HAS_PTHREAD=0
```
seantalts: why is this step suddenly necessary? Does Jenkins have to change?

syclik: Just hunted it down. It has to do with the os_linux in the math library setting it for pretty much every compiler option there. It trickles down from `CC`. Let me figure out a different way around it.

syclik:
Issues in other repos for CXX / CC:
- math already exists: stan-dev/math#617
- stan (created a new issue): stan-dev/stan#2444

syclik:
@seantalts, just made some changes. To address the three things:
- There are now issues in Stan and Math for using CXX; I didn't add one for CmdStan because it's fine with CXX now.
- I added a comment to the very top of the makefile. Feel free to change or make suggestions.
- for the installation.tex, I didn't change it just because it's described in the subsection right before. If you have any suggestions on how to make that better, want to do that?

syclik:
I also updated `make/tests` to include the gtest-related pthread flag for linux. I left a fixme so we remember to remove it later.

PR #604, opened 2018-01-22
Bugfix/issue 602 adapt stepsize

bbbales2:
#### Submisison Checklist

#### Summary:

Adds more tests for custom specified metrics and also adds an error that gets thrown if adaptation is enabled with num_warmup=0

To reviewer: Am I putting that check in the right place? It's right here: https://github.com/bbbales2/cmdstan-rus/blob/95cb7995d25c307f6054f53ad4675e5cbe80c08e/src/cmdstan/command.hpp#L255. Doesn't seem to be a whole lot of other checking going on in command.hpp.

#### How to Verify:

Build the bernoulli example and then try to run it with:

```
./bernoulli method=sample num_samples=1 num_warmup=0 adapt engaged=1 algorithm=hmc engine=nuts stepsize=1e-7 data file=bernoulli.data.R output file=output1.csv
```

And you should get the message:

"The number of warmup samples (num_warmup) must be greater than zero if adaptation is enabled."

And the program should exit

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company): University of California, Santa Barbara

### DIFF
--- a/src/cmdstan/command.hpp
+++ b/src/cmdstan/command.hpp
@@ -252,7 +252,10 @@ namespace cmdstan {
         double stepsize = dynamic_cast<real_argument*>(hmc->arg("stepsize"))->value();
         double stepsize_jitter= dynamic_cast<real_argument*>(hmc->arg("stepsize_jitter"))->value();
 
-        if (engine->value() == "nuts" && metric->value() == "dense_e" && adapt_engaged == false && metric_supplied == false) {
+        if (adapt_engaged == true && num_warmup == 0) {
+          info("The number of warmup samples (num_warmup) must be greater than zero if adaptation is enabled.");
+          return_code = stan::services::error_codes::CONFIG;
+        } else if (engine->value() == "nuts" && metric->value() == "dense_e" && adapt_engaged == false && metric_supplied == false) {
           int max_depth = dynamic_cast<int_argument*>(dynamic_cast<categorical_argument*>(algo->arg("hmc")->arg("engine")->arg("nuts"))->arg("max_depth"))->value();
           return_code = stan::services::sample::hmc_nuts_dense_e(model,
                                                                  init_context,
--- a/src/docs/cmdstan-guide/running.tex
+++ b/src/docs/cmdstan-guide/running.tex
@@ -810,8 +810,9 @@ The \code{metric\_file} option can be used with and without adaptation enabled.
 If adaptation is enabled, the provided metric will be used as the initial
 guess in the adaptation process. If the initial guess is good, then adaptation
 should not change it much. If the metric is no good, then the adaptation will
-override the initial guess. An example of running the sampler with adaptation
-enabled without specifying a stepsize is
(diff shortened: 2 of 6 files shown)


bbbales2:
Now that I think about it, an alternate and probably better solution to this is to just require num_warmup > 0.

wds15:
Are u sure about num_warmup>0 constraint?

Would it still be possible then to restart sampling from exactly the same draw+warmup state? I think this would not be possible then and at least to me that is a desirable feature.

bob-carpenter:
We can start from the same draw plus warmup, but we can't start from the same RNG state because the RNG state isn't translatable back into a seed.  We'd have to rethink our core RNG to make saving and restarting the RNG possible.  It'd be easy if it kept track of the number of calls it made because it has an efficient advance-over-the-next-N-draws method.

bbbales2:
@wds15 If you want to restart something with known mass matrix n' such, adapt engaged=0 is the way to go. adapt engaged=0 will skip warmup (so it doesn't matter what it's set at).

bbbales2:
Jenkins, retest this please

wds15:
ah, ok. I think some hints in the documentation would be great in this regard. We should emphasize that restarting needs engaged=0 and the stepsize set specifically (and that chains don't give us the same output due to rng state). If you like I give that a try or would you like to work on that as well?

bbbales2:
The other thing that makes sense I think is that setting num_warmup=0 would just set adapt engaged=0 as well. I guess this pull is just waiting on whoever is the cmdstan Czar to decide if this should be handled with:

1. Error thrown if num_warmup=0 & adapt engaged=1, or
2. Require num_warmup > 0, or
3. Force adapt engaged=0 if num_warmup=0

Once that's decided the docs are straightforward.

I dunno the philosophy in how these sorts of edge cases are handled, but it seems like there'd be a Best Way to do it so that it'd trickle out to the other interfaces without being difficult to handle. 1 & 3 seem a little bandaidy in implementation, 2 is probably the cleanest.

betanalpha:
Turning on adaptation with engaged=1 initializes the
adaptation immediately regardless of how many warmup 
iterations are run.  This is the behavior independent of
the iterations run is expected and desired behavior.

Currently the initialization sets the step size to the value 
of the "algorithm hmc stepsize=blah” argument which 
defaults to 1.  Unless the user provides a different step
size there’s nothing more the adaptation routine can do
to better initialize.

wds15:
Should we run this by the stan-dev team on Thursday?

I like the proposal to set engaged=0 whenever num_warmup=0.

bbbales2:
Yeah, sounds like a good idea

bbbales2:
@mitzimorris I added a tiny bit of documentation to the manual to clarify that num_warmup=0 isn't valid if adapt engaged=1. Otherwise this pull does what we talked about in the meeting, so it's ready to review.

Thanks!

mitzimorris:
reviewed and approved.  apologies for letting this languish.

Issue #658, opened 2018-12-01
Random seed of 0 accepted, yields same results as random seed 1


riddell-stan:
#### Summary:

According to ``help-all`` a random seed of 0 is invalid. Not only is it accepted, it seems to be equivalent to specifying a random seed of 1.

#### Reproducible Steps:
```bash
examples/bernoulli/bernoulli sample random seed=0 data file=examples/bernoulli/bernoulli.data.R output file=output-seed-0.csv
examples/bernoulli/bernoulli sample random seed=1 data file=examples/bernoulli/bernoulli.data.R output file=output-seed-1.csv
diff output-seed-0.csv output-seed-1.csv
```

#### Current Version:
v2.18.0

riddell-stan:
Happy to submit a PR if someone tells me what the desired behavior is.

mitzimorris:
I see the fix that's needed - the cmdstan argument parser allows any int value.
the fix is to add function "is_valid" to file 'cmdstan/arguments/arg_seed.hpp'
will do ASAP.

mitzimorris:
fixed by https://github.com/stan-dev/cmdstan/pull/695

PR #695, opened 2019-04-25
Bugfix/658 validate random seed

mitzimorris:
#### Submisison Checklist

#### Summary:
added checks to random seed argument and corresponding unit tests

#### Intended Effect:
make actual behavior match advertised behavoir

#### How to Verify:
unit tests

#### Side Effects:
n/a

#### Documentation:
n/a

#### Copyright and Licensing

### DIFF
--- a/src/cmdstan/arguments/arg_seed.hpp
+++ b/src/cmdstan/arguments/arg_seed.hpp
@@ -13,7 +13,7 @@ namespace cmdstan {
       _description = "Random number generator seed";
       _validity = "integer > 0 or -1 to generate seed from sys clock time";
       _default = "-1";
-      _default_value = 0 - 1;  // -1 as unsigned int
+      _default_value = -1;
       _constrained = true;
       _good_value = 18383;
       _bad_value = -2;
--- a/src/cmdstan/arguments/arg_seed.hpp
+++ b/src/cmdstan/arguments/arg_seed.hpp
@@ -11,12 +11,18 @@ namespace cmdstan {
     arg_seed(): u_int_argument() {
       _name = "seed";
       _description = "Random number generator seed";
-      _validity = "seed > 0, if negative seed is generated from time";
+      _validity = "integer > 0 or -1 to generate seed from sys clock time";
       _default = "-1";
-      _default_value = -1;
-      _constrained = false;
+      _default_value = 0 - 1;  // -1 as unsigned int
+      _constrained = true;
       _good_value = 18383;
+      _bad_value = -2;
       _value = _default_value;
+    }
+
+    bool is_valid(unsigned int value) {
+      return (value > 0 && value <= std::numeric_limits<int>::max())
+        || (value == _default_value);
     }
   };
 
--- a/src/cmdstan/arguments/arg_seed.hpp
+++ b/src/cmdstan/arguments/arg_seed.hpp
@@ -6,12 +6,12 @@
 namespace cmdstan {
 
 
-  class arg_seed: public u_int_argument {
+  class arg_seed: public int_argument {
   public:
-    arg_seed(): u_int_argument() {
+    arg_seed(): int_argument() {
       _name = "seed";
       _description = "Random number generator seed";
-      _validity = "integer > 0 or -1 to generate seed from sys clock time";
+      _validity = "integer > 0 or -1 to generate seed from system time";
       _default = "-1";
       _default_value = -1;
       _constrained = true;
@@ -20,9 +20,8 @@ namespace cmdstan {
       _value = _default_value;
     }
 
-    bool is_valid(unsigned int value) {
-      return (value > 0 && value <= std::numeric_limits<int>::max())
-        || (value == _default_value);
+    bool is_valid(int value) {
+      return value > 0 || value == _default_value;
     }
   };
 
--- a/src/test/interface/arguments/argument_configuration_test.cpp
+++ b/src/test/interface/arguments/argument_configuration_test.cpp
@@ -67,53 +67,53 @@ TEST_F(StanGmArgumentsConfiguration, TestMethod) {
 
   std::vector<cmdstan::argument*> valid_arguments;
   valid_arguments.push_back(new cmdstan::arg_method());
-  
+
   cmdstan::argument_probe probe(valid_arguments);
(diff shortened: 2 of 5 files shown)


seantalts:
This fixes #695 right?

seantalts:
This fixes #658 right?

mitzimorris:
this fixes #695

mitzimorris:
this fixes https://github.com/stan-dev/cmdstan/issues/658

mitzimorris:
@syclik - can you take a look at this?  the fix here is adding a check to the random seed argument.  simplified the unit test to get to the essential w/r/t the order of the method and random seed configs.  added unit tests to make sure that seed values behave as advertised.

src/cmdstan/arguments/arg_seed.hpp
```diff
@@ -11,13 +11,19 @@ namespace cmdstan {
     arg_seed(): u_int_argument() {
       _name = "seed";
       _description = "Random number generator seed";
-      _validity = "seed > 0, if negative seed is generated from time";
+      _validity = "integer > 0 or -1 to generate seed from sys clock time";
```
syclik: Rather than "sys clock time" could we just write that out? I don't know if casual users would understand `sys clock`... actually I googled it and couldn't find hits.

Maybe replace that with "system time"? If someone is confused and they search for it, the first hit I find is wiki: https://en.wikipedia.org/wiki/System_time; and I think that's correct enough.

src/test/interface/arguments/argument_configuration_test.cpp
```diff
@@ -838,81 +838,20 @@ TEST_F(StanGmArgumentsConfiguration, TestRandomWithMethod) {
...
-      
-    }
-    
-  }
-  
-  for (size_t i = 0; i < valid_arguments.size(); ++i)
-    delete valid_arguments.at(i);
-  
+
+  // check good
```
syclik: It looks like the old tests were mistakenly removed. Was this intentional?

Maybe the test you have below should be moved (or is already tested) in `command_test.hpp`?

mitzimorris:
> @mitzimorris, the change itself looks good. I'm just confused on the removal of the test. If there's a good reason for it, then I'm happy to approve it with the tests removed. When I look at the diff, it looks like you've shortcutted most of the test to test the new changes (that's good), but it seems like it was meant to be a temporary change while debugging.

I couldn't get the test framework to work with full generality.  if you understand how these templates work, that would be great, because I couldn't see why the changes I made should have had any impact on the test - I agree that the unit tests shouldn't have to be changed because there might be something that they're testing that I don't understand.

I think this unit test is designed to check that the "random seed=xxx" argument can come before or after the "method ...." set of arguments.    which is what the simple version does, without testing the argument parser in full generality.  happy to discuss further.

syclik:
@mitzimorris, I'm sorry -- I completely missed your comment.

I don't understand how these templates work, but maybe we should revert the tests and see where it fails? What do you think we should do next?

mitzimorris:
go for it!

mitzimorris:
it wasn't clear that the template for testing "random" had ever been called as part of a test, so not sure it ever worked.

syclik:
I'm making some quick mods to the branch. I'm testing now... hopefully
it'll work.

syclik:
I don't know what's going on and it's late. I'm calling it for now; I'll try to look at it again soon.

syclik:
@mitzimorris, I think I figured it out. It was ugly getting it to work; there's a mix of templates + explicit template instantiations + object oriented programming + pointers all wrapped into one thing. But it's fixed.

I also had to handle the exception on `boost::lexical_cast`; thanks for putting in that test case.

mitzimorris:
> there's a mix of templates + explicit template instantiations + object oriented programming + pointers all wrapped into one thing

now that you see what's going on, do you agree with my assessment of what the test is trying to do, i.e.:

> check that the "random seed=xxx" argument can come before or after the "method ...." set of arguments.

and if so, why not go with a simpler test and get rid of obviously unmaintainable code?

syclik:
sorry, some more clarification please? what is...
- "the test"?
- what's your assessment of what it's trying to do?
- what would be a simpler test?

I agree that this is unmaintainable, but I don't think removal of tests is going to make it more maintainable. I think we should go with a different mechanism for handling command-line arguments completely.

syclik:
(that said, this test is really hard to read)

syclik:
@mitzimorris, I was able to fix the test. The same error was actually all over the test file, so it's a wonder the test actually worked at all. I think your assessment of the worth of the test long-term is correct: it may be buggy and not actually testing the right thing. And it's hard to even verify what's going on.

That said, I'm still hesitant to remove tests without a replacement. I didn't think I removed your new tests... if I did, we can bring them back.

mitzimorris:
> I think we should go with a different mechanism for handling command-line arguments completely.

I completely agree and would be happy to take this one on.

syclik:
Let's work through a design doc? I've got ideas, but they're not flushed
out yet. Or we can start on discourse. Your choice.

mitzimorris:
a design doc would be great.
there's now a 'design-docs' repo.

bob-carpenter:
One of the motivations for keeping arguments independent is that it makes validation of arguments much much simpler.  I think this has to happen for all args on the Python side in order to push out error messages using the right terminology.  Otherwise, we'll just see CmdStan error messages.

I'm in favor of writing a simple library of test functions like we have on the C++ side.   For instance, we'll need to test that metrics are positive definite, that the random initialization scale is positive, that the random seed's an integer, etc.

mitzimorris:
@syclik are you OK with tests for this?  would be nice to merge for release (if possible)

mitzimorris:
@syclik - are you OK with the tests are they are?  if so, please sign off on the review and I will merge this - hoping to get it in for the next release.

mitzimorris:
@syclik - many thanks - will try to do better w/r/t putting essentials all in same message.

syclik:
Thanks! Every little bit to give the right context helps. If you can’t
tell, we’re both working across a few issues across different repos
simultaneously. It’s not always easy to keep track of the thread while
juggling a bunch more.

PR #700, opened 2019-05-10
[Jenkins] Try running interface tests in parallel on Jenkins

seantalts:
This used to not work for some reason but it might now that there was that big makefile refactor. Fingers crossed!


seantalts:
Yay! Looks like it used to take ~14 minutes on Windows and now it takes ~3. I think this ability was fixed by the makefile refactor, @syclik  :)

syclik:
Awesome!

PR #706, opened 2019-05-23
Interface tests now run on Windows, Mac and Linux.

serban-nicusor-toptal:
#### Submisison Checklist

#### Copyright and Licensing

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -3,15 +3,21 @@ import org.stan.Utils
 
 def utils = new org.stan.Utils()
 
-def setupCXX(CXX = env.CXX) {
+def setupCXX(failOnError = true, CXX = env.CXX) {
     unstash 'CmdStanSetup'
-    writeFile(file: "make/local", text: "CXX = ${CXX}\n")
+    errorStr = failOnError ? "-Werror " : ""
+    writeFile(file: "make/local", text: "CXX=${CXX} ${errorStr}")
 }
 
-def runTests(String prefix = "") {
+def runTests(String prefix = "", String testPath) {
     """ make -j${env.PARALLEL} build
-  ${prefix}runCmdStanTests.py -j${env.PARALLEL} src/test/interface
+      ${prefix}runCmdStanTests.py -j${env.PARALLEL} ${testPath}
     """
+}
+
+def deleteDirWin() {
+    bat "attrib -r -s /s /d"
+    deleteDir()
 }
 
 pipeline {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -3,17 +3,23 @@ import org.stan.Utils
...
-def setupCXX(CXX = env.CXX) {
+def setupCXX(failOnError = true, CXX = env.CXX) {
     unstash 'CmdStanSetup'
-    writeFile(file: "make/local", text: "CXX = ${CXX}\n")
+    errorStr = failOnError ? "-Werror " : ""
+    writeFile(file: "make/local", text: "CXX=${CXX} ${errorStr}")
 }
 
-def runTests(String prefix = "") {
+def runTests(String prefix = "", String testPath) {
```
seantalts: we don't need to pass the path in here :P

Jenkinsfile
```diff
@@ -3,17 +3,23 @@ import org.stan.Utils
 
 def utils = new org.stan.Utils()
 
-def setupCXX(CXX = env.CXX) {
+def setupCXX(failOnError = true, CXX = env.CXX) {
     unstash 'CmdStanSetup'
-    writeFile(file: "make/local", text: "CXX = ${CXX}\n")
+    errorStr = failOnError ? "-Werror " : ""
```
seantalts: we don't need this

seantalts:
Thanks!

Build failed, maybe it needed the newline? Oh, I think you can't have default arguments that look in the environment, because it doesn't get populated when the script is first imported. If that's right you'll need to set the default to something like null and then check if it's null and set it to env.CXX or whatever.

serban-nicusor-toptal:
I think default arguments that look into `env.` should work just fine.
It's like that in Stan Jenkinsfile for the Windows part.

seantalts:
Where do you see that?
To clarify I mean this part:
https://github.com/stan-dev/cmdstan/pull/706/files#diff-58231b16fdee45a03a4ee3cf94a9f2c3R6

serban-nicusor-toptal:
My bad, it was actually in cmdstan:
https://github.com/stan-dev/cmdstan/blob/develop/Jenkinsfile#L6

seantalts:
Ah okay, maybe that does work now (though I swear I had problems with it a few years ago). Is it possible the Linux EC2 nodes don't have GCC env var defined?

seantalts:
oops, no it failed on a Mac. hmm.

serban-nicusor-toptal:
Hmh, the Macs do not have the `GCC` var set.

seantalts:
Oh right, we want to run on Clang (e.g. CXX) on Macs and GCC on Linux only.

Jenkinsfile
```diff
@@ -48,30 +53,50 @@ pipeline {
...
+                    }
                 }
-                stage('Non-windows interface tests') {
-                    agent any
+
+                stage('Linux interface tests') {
+                    agent {label 'linux'}
                     steps {
-                        setupCXX()
+                        setupCXX("${env.GCC}")
```
seantalts: don't need to string interpolate this, you can just pass env.GCC here

serban-nicusor-toptal:
Ah! Give me one sec to move Macs to CXX

Jenkinsfile
```diff
@@ -48,30 +53,50 @@ pipeline {
...
-                    post { always { deleteDir() }}
+                    post { 
+                        always { 
+
+                            recordIssues id: "Windows",
+                            name: "Windows interface tests",
+                            enabledForFailure: true,
+                            aggregatingResults : true,
+                            tools: [
+                                gcc4(id: "Windows_gcc4", name: "Windows interface tests@GCC4"),
```
seantalts: This `id` thing is very cool, did not know how to do that before and we got kind of messy results from it. Thanks :D

seantalts:
Need to replace "&" with "&&" in label AND operation

PR #710, opened 2019-06-16
add Intel TBB

wds15:
#### Submisison Checklist

#### Summary

This adds the TBB targets to the makefiles. This PR needs to be merged with the Stan-math PR https://github.com/stan-dev/math/pull/1180 and the stan PR https://github.com/stan-dev/stan/pull/2769

#### Intended Effect

Make the Intel TBB available to stan.

#### How to Verify

Tests will link against the tbb, tbbmalloc & tbbmalloc_proxy libraries.

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company): Sebastian Weber


wds15:
@serban-nicusor-toptal How do the performance tests work when I use the "build with parameters" facility to use non-default stan and stan-math refs to build this PR? Moreover, is it possible to instruct the performance suite to test on macOS specifically (and maybe Linux)?

serban-nicusor-toptal:
@wds15 Hey, 

It will get the base branch `develop`,
Checkout cmdstan and submodules from that branch/commit hash in `build with parameters`.
Then run tests on develop with latest stable submodules versus the branches/hashes in `build with parameters`.

https://github.com/stan-dev/performance-tests-cmdstan/blob/master/compare-git-hashes.sh

---

I am currently working on a custom branch in git and job for Jenkins to allow running the tests on mac/win/linux with more arguments for ease of customization.  
I will add the possibility to run on a custom set of oses.

See: 
- https://github.com/stan-dev/performance-tests-cmdstan/tree/custom
- https://jenkins.mc-stan.org/job/CmdStan%20Performance%20Tests/job/Custom/

Still needs a bit of refactoring to work on win.

mitzimorris:
is this still status WIP?

wds15:
Yes...the tbb still needs to have settlement on the license

.travis.yml
```diff
@@ -50,20 +51,24 @@ matrix:
       env:
         - MATRIX_EVAL="CXX=clang++-3.8"
           CXXFLAGS="-stdlib=libc++"
+          CXX_TYPE="clang"
```
rok-cesnovar: So for mpicxx this has to be set. But it does not need to be set here right?

.travis.yml
```diff
@@ -50,20 +51,24 @@ matrix:
...
           PARALLEL=2
     - <<: *linux_gcc
       env:
         - MATRIX_EVAL="CXX=g++-4.9"
+          CXX_TYPE="gcc"
```
rok-cesnovar: Or this?

.travis.yml
wds15: ah, I see. Probably not. It depends on wether our makefiles correctly figure out the CXX type from the CXX variable.

.travis.yml
wds15: same thing here. Let me check.

rok-cesnovar: Ah, I now see that these were in .travis.yml which has since been deleted. Sorry for wasting your time here :)

wds15: NP... it was a fast fix for me.

wds15:
Right now I am outputting on windows a message about adding the tbb library path to the PATH variable... but we could also just create a wrapper `.bat` file for each model on Windows. The wrapper will have the hard-coded TBB path in it, set the PATH variable and then start the stan program. This way we also get to control which TBB is being used on the system as we can control to put our TBB in the first place of the PATH. That wrapper should be easy to write.

wds15:
For reference, here is the windows search policy for dynamic libraries:

https://docs.microsoft.com/en-us/windows/win32/dlls/dynamic-link-library-search-order#standard-search-order-for-desktop-applications

rok-cesnovar:
> and then start the stan program.

By this you mean stan compiler? That sound like a nice convenience for Windows users and I very much approve. 

But honestly, I dont feel this convenience .bat file is necessary in this initial TBB PR. If you feel that it will be easy to write then go ahead. If not, we can do this separately after this initial batch of TBB PRs get in.

rok-cesnovar:
Or did you mean that a user would run their models through this .bat file?

wds15:
Yes, users would call the .bat file instead of the .exe file.

I would generate the .bat with the makefile.

So calling the Bernoulli example becomes

`bernoulli.bat sample ... `

that's it. The bernoulli.bat just wraps the exe call which it will precede with a reset PATH variable.

rok-cesnovar:
This means that PATH is set each time a model is run? Or am I misunderstanding.

What about just setting PATH on `make build` of CmdStan?

rok-cesnovar:
Ah, I see. I missed that setx only works if you run with administrator, which is a pain.

So a user would have the option of: 
- setting the path manually and running an exe file the way it was done until now
- run the model through a .bat file (basically the way it was done now but exe becomes bat)

@mitzimorris @ahartikainen On update of cmdstanpy to cmdstan 2.21 you will most likely want to run the model through the .bat file for convenience of Windows users. Everything else stays the same. Want us to create an issue on your repository? I imagine it wont be a huge change.

ahartikainen:
For CmdStanPy we can edit the $PATH for each run if needed.

tbh I don't like .bat files and all the edgecases it will have (probably something we don't currently think like calling it from cmd.exe vs powershell etc), but that is my personal opinion :)

rok-cesnovar:
Yes, you will have to add the path to TBB of each run. If you have easier ways of doing that for cmdstanpy that is great.

I agree that .bat files are not perfect, but powershell scripts are probably worse :) The users will always have the option of manually adding TBB to PATH and doing everything as they did before. Its just a convenience add-on.

ahartikainen:
Ok, sounds good.

In python we have

    import os
    env = os.environ.copy()
    env["PATH"] = "{};{}".format(os.path.abspath("path/to/lib"), env["PATH"])

and then when we call subprocess

    subprocess.Popen(..., env=env)

rok-cesnovar:
Great, thanks for the quick response! Will make an issue once we hash out all the details in cmdstan and Stan Math.

wds15:
So is the conclusion that we do not want a .bat file?

Maybe quickly touch base on this in the Stan meeting?

rok-cesnovar:
The conclusion is that if we can make .bat work without much complication we should do it. Lets try to talk about that at the Stan meeting, hopefully I make it home in time...

mitzimorris:
are we proposing that in the future, CmdStan generates .bat instead of .exe files on Windows?

this is problematic in the case where users have built their own lightweight wrappers around CmdStan - and I suspect that many people have done this - they update to the next release and suddenly everything breaks.  please let's not do this.

rok-cesnovar:
> are we proposing that in the future, CmdStan generates .bat instead of .exe files on Windows?

No, .exe stays. 

.bat would be added as a convenience. Because with the addition of TBB the user will  need to modify the PATH variable in order to link with TBB (Windows does not support dynamic linking with relative paths).

So a user would have the option of:

- setting the path manually, everything else stays the as it was pre 2.21
- run the model through a .bat file; .bat takes cares of setting the PATH and running the .exe

mitzimorris:
I took a quick look at the makefile, does this mean that for Windows users who aren't using TBB, the variable TBB_TARGETS will be null and nothing happens?  and it's only folks who want to use TBB that end up with both .EXE and .BAT files?

wds15:
No. The TBB will be mandatory - so everyone has to do go through this. The current proposal is to have the .bat in addition to the .exe. Windows users either use the .bat from now on or they stay with .exe, but set the PATH variable (or they could actually also copy the tbb dlls into system locations).

mitzimorris:
this will be very confusing for windows users - now they've got twice as many files lying around.

ahartikainen:
Hmm, is there a way to pack tbb.dll to exe and use that if tbb.dll is not found in the $PATH

wds15:
It's possible to write windows specific code in C++ which will modify the search path and then you also need to load the libraries "manually" in the code. So this means to add windows specific C++ code into the codebase. That's really not nice.

So either

- one more file, a .bat file
- the PATH variable must be set

are really the preferred options here from my view. I don't see what's confusing about calling a bat file instead of a .exe file?

mitzimorris:
>  I don't see what's confusing about calling a bat file instead of a .exe file?

it breaks backwards compatibility.

my hunch is that CmdStan is being called from scripts that users have developed to run pipelines on their systems.  they have no way of knowing when these scripts need to be changed.  furthermore, if they're done a really good job on the scripting, the people using the scripts are not the same people who wrote them, so when the scripts break, all they know is that "Stan is broken".

given this, changing from .exe to .bat files is problematic.

wds15:
The how about (in addition) a

`make install-tbb`

target which will copy the tbb*.dll's into the windows system directories?

Or maybe there is even a magic windows command which sets globally the PATH variable for the user? I am not a windows expert here.

"Breaking backward" compatibility is a killer argument - usually. I do disagree in this matter here though. People can still just define the PATH variable and then they are set to use the .exe just as before. We are making additional requirements on the environment, which is fine as long as we document it.

mitzimorris:
> The how about (in addition) a
> 
> `make install-tbb`
> 
> target which will copy the tbb*.dll's into the windows system directories?

so that way .exe files will just work as before?

I like this - expecially since this could be added to CmdStanPy's "install CmdStan" script.

regarding "backwards compatibility" - at this point we've got a lot of users and a lot of technical debt, which makes steering the project increasingly difficult.  I feel your pain and frustration.

ahartikainen:
If it doesn't need admin rights.

Is there a location on Windows which is under user and in $PATH?

src/docs/cmdstan-guide/getting-started.tex
```diff
@@ -15,7 +15,11 @@ \section{Installation}
   download.
 \item The \code{make} utility program. \\
   This is not strictly necessary, but will make the build process
-  easy. The rest of the documentation assumes \code{make} is available.
+  easy. On the Windows platform the \code{mingw32-make} variant is
```
rok-cesnovar: Just say "On Windows \code{mingw32-make} is needed to build \CmdStan. This is a \code{make} variant and can be installed as part of Rtools ...".

src/docs/cmdstan-guide/installation.tex
```diff
@@ -644,6 +672,10 @@ \subsection{GNU Make Build Tool}
 \item Tested Versions: 3.81 (Mac OS X), 3.79 (Windows 7)
 \end{itemize}
 %
+On Windows the \code{mingw32-make} variant must be used, since the
```
rok-cesnovar: "On Windows \code{mingw32-make} must be used, as it is a requirement to build Intel TBB. This variant of make is available..."

rok-cesnovar:
@wds15 I pushed the script to set the path on Windows to the branch. 

`make install-tbb` now runs this .bat file that sets the path.

I will just write down why I had to do it this way, if anyone else wants to improve this in the future and does not fall in the same pitfalls as me:
1. we want to set the path in the users space, not systems, that would require admin privileges
2. however %PATH% returns the concatenated values of path in the system and user space (YAY!)
3. `SETX Path TBB_PATH;%PATH%` is therefore not an option as we would insert the entire system path to the users variable which is a no no
4. So how do you get the users path?
`reg query HKEY_CURRENT_USER\Environment /v path` returns something like

```
HKEY_CURRENT_USER\Environment
    path    REG_SZ    C:\Python27;C:\Rtools\mingw_64\bin;
```
5. We then grep and cut that value out and save it to a temporary file. 

`reg query HKEY_CURRENT_USER\Environment /v path | sed -n 's/.*REG_SZ[ ]*//p' > temp.txt`

Sed is part of GNU tools which RTools installs so I guess we can use it here.

6. Read back the file and set the users path plus the path to the TBB (%~dp0 is the location of the .bat file. Yep.). Then remove the temp file.

set /p VAR=<temp.txt
SETX Path %~dp0\lib\tbb;%VAR%
rm temp.txt

I was not able to put the file in the make folder as that did not work with the make system. It kept trying to find rules to build it or whatever.

The only question is do we want to have this as separate "make install-tbb" or just call it at the end of "make build" for Windows users which I believe would also be just fine, except if anyone wants to use their own TBB install. That is your call Sebastian.

wds15:
@rok-cesnovar thanks for this work on PATH. I tried it and it failed to work for me as I got a slightly different output from the registry call. Using `sed` is in principle fine, but that would require documentation on that we need this.

Luckily I found solution here: https://stackoverflow.com/questions/19287379/how-do-i-add-to-the-windows-path-variable-using-setx-having-weird-problems

Now there is no more need to create a temporary file nor do we need `sed`.

My preference here is to inform users what they should do after build completes. So the build command prints what users need to do after having gone through `mingw32-make build`. 

Could you try things out to the point where you can run the bernoulli example? Thanks.

I will update the doc and then this is good, I think.

install-tbb.bat
```diff
@@ -0,0 +1,9 @@
+@ECHO off
+echo Permanently setting the PATH user environment variable:
+
+for /F "tokens=2* delims= " %%f IN ('reg query HKCU\Environment /v PATH ^| findstr /i path') do set OLD_SYSTEM_PATH=%%g
+setx Path %~dp0stan\lib\stan_math\lib\tbb;%OLD_SYSTEM_PATH%
```
rok-cesnovar: If built with cmdstan tbb is build to `cmdstan\lib\tbb` not `cmdstan\stan\lib\stan_math\lib\tbb`. Either fix this path here or make tbb build to the stan_math folder. I would prefer the latter.

rok-cesnovar: Other than that I can confirm this works. I like that its much more windows-like than my sed thing.

wds15: For me the TBB gets build in cmdstan/stan/lib/stan_math/lib/tbb which is the correct location, I think. It has to be under stan_math as it is part of stan_math (and the makefiles for it live there).

wds15:
I just did

`git checkout origin/develop stan`

and it did not change anything as it looks to me. So I am not sure what I should update wrt to the submodule.

install-tbb.bat
rok-cesnovar: I will do a clean clone to double check.

rok-cesnovar:
If you look at the changes in the PR it looks as we are changing the stan submodule.which we should not.

install-tbb.bat
rok-cesnovar: On windows I get the following:
```
NOTE: Please add C:/Users/Rok/Desktop/cmdstan/lib/tbb to your PATH variable.
You may call

mingw32-make install-tbb
```

On Ubuntu its the same thing. Are you sure its where it gets build on your system? I agree that stan_math is the right place.

wds15: argh.. found it. Fixing it on the math tbb-lib branch. Sorry.

rok-cesnovar: Great. This isnt a problem with this PR so resolving.

wds15:
Cool. Let's also wait for the pipeline of stan-math PR-1180 to turn green before we merge this. Then we know that all will work out.

(I really hope that this time Jenkins is stable, the thing started to give again flaky behavior here and there)

rok-cesnovar:
Did you setup 1180 to the TBB stan and cmdstan branches? Yeah, it would be great if we had that green light before.

I would then merge these three in quick succession and then deal with 1376. Then things will get easier.

Will handheld Jenkins if needed, no problem :)

rok-cesnovar:
On the bright side we have come to the point where the only thing keeping us from having TBB in is having to wait for the tests to finish. I am psyched :)

wds15:
Me too ... looking forward to this.

So yes, I started the Jenkins for PR 1180 of stan-math using the "build with parameters" facility. That lets you test the 1180 with the respective upstream branches which match it.

I will update the PR test for 1180. If you don't mind I will take out the performance things there as these are documented in 1376.

BTW... the very last one to merge is the changes currently on the branch `feature/intel-tbb-init
` in cmdstan. That's a super easy one which can do at the very end (to ensure proper interface initialization).

rok-cesnovar:
> I will update the PR test for 1180. If you don't mind I will take out the performance things there as these are documented in 1376.

You mean the PR text? Yes, please do that. Just reference to 1376 for performance stuff. 1180 doesnt really change anything outside of Mac's use of tbbmalloc in terms of performance.

makefile
```diff
@@ -152,14 +152,29 @@ build-mpi: $(MPI_TARGETS)
 	@echo '--- boost mpi bindings built ---'
 
 .PHONY: build
-build: bin/stanc$(EXE) bin/stansummary$(EXE) bin/print$(EXE) bin/diagnose$(EXE) $(LIBSUNDIALS) $(MPI_TARGETS) $(CMDSTAN_MAIN_O)
-	@echo ''
+build: bin/stanc$(EXE) bin/stansummary$(EXE) bin/print$(EXE) bin/diagnose$(EXE) $(LIBSUNDIALS) $(MPI_TARGETS) $(TBB_TARGETS) $(CMDSTAN_MAIN_O)
+	@echo ''
+	@if [ "$(OS)" == "Windows_NT" ]; then \
```
rok-cesnovar: Argh, I found an ugly warning that doesnt really prevent anything, the thing still builds. Its just ugly.
On ubuntu I get

`/bin/sh: 1: [: Linux: unexpected operator`

Mind changing this to 

```
.PHONY: build
build: bin/stanc$(EXE) bin/stansummary$(EXE) bin/print$(EXE) bin/diagnose$(EXE) $(LIBSUNDIALS) $(MPI_TARGETS) $(TBB_TARGETS) $(CMDSTAN_MAIN_O)
	@echo ''
ifeq ($(OS),Windows_NT)
		echo 'NOTE: Please add $(TBB_BIN_ABSOLUTE_PATH) to your PATH variable.';
		echo 'You may call';
		echo '';
		echo 'mingw32-make install-tbb';
		echo '';
		echo 'to automatically update your user configuration.';
endif
	@echo '--- CmdStan v$(CMDSTAN_VERSION) built ---'
```

makefile
```diff
@@ -152,14 +152,29 @@ build-mpi: $(MPI_TARGETS)
...
 	@echo '--- CmdStan v$(CMDSTAN_VERSION) built ---'
 
 ifeq ($(CXX_TYPE),clang)
 build: $(STAN)src/stan/model/model_header.hpp.gch
 endif
 
+.PHONY: install-tbb
+install-tbb: $(TBB_TARGETS)
+	@if ! [ "$(OS)" == "Windows_NT" ]; then \
+		exit 0; \
```
rok-cesnovar: And this to 

```
ifeq ($(OS),Windows_NT)
		exit 0;
endif
```

Sorry I missed this one before.

makefile
wds15: sure.

makefile
wds15: ok. much nicer.

wds15:
Now the final message looks nicer.

rok-cesnovar:
Great, thanks. 

Cmstan and Stan PRs are ready and waiting. When/if 1180 passes we can then merge all three. If all goes well and Jenkins plays nicely we should be there in around 3 hours (1h for distribution test and 1h30min for upstream). I will definitely still be at my computer then to merge. Should I just do it then? If all goes green I think we should.

And we then handle 1376 tomorrow once that passes? And the final Cmdstan PR to init the number of threads.

wds15:
Sure - go ahead once 1180 turns green. Awesome.

PR #723, opened 2019-08-14
Bugfix 722 Make the default seed produce varying results again

### LINKED ISSUE #722: Default Random Seed Doesn't Work As Expected
#### Description:

Default random seed of -1 returns consistent results for each evaluation and not the varying results expected by the documentation and tests.  This is causing upstream tests to fail blocking development of the core Stan code.

#### Reproducible Steps:

./runCmdStanTests.py src/test/interface/command_test.cpp

fails for the v2.20.0 tag [OS X 10.12.6, Apple LLVM version 9.0.0 (clang-900.0.39.2)].

The behavior can also be reproduced by running a model, say
```
parameters {
  real x[50];
}

model {
  x ~ normal(0, 1);
}
```
with default settings and examining the resulting output.csv files.

#### Current Output:

Failing tests.

#### Expected Output:

Not failing tests.

#### Additional Information:

I can't find any documentation in the Boost online docs or the 1.69 code currently in Stan that states that seed=-1 yielding random behavior is supported so I can't verify expected behavior.

#### Current Version:
v2.20.0

rok-cesnovar:
#### Submisison Checklist

#### Summary:

This fixes #722 by casting the calculated value for the random seed to an unsigned int. The time difference used for the random seed was previously assigned to an integer which produced a negative seed. This caused consistent results with the default seed of -1 on all runs.

The other solution to the same problem would be to replace 
`(boost::posix_time::microsec_clock::universal_time() - boost::posix_time::ptime(boost::posix_time::min_date_time)).total_milliseconds();` with something that returns only positive ints

#### Intended Effect:

The default seed of -1 should produce varying results.

#### How to Verify:

Run `command_test.hpp`

#### Side Effects:

/

#### Documentation:

/

#### Copyright and Licensing

Rok Češnovar, Univ. of Ljubljana

### DIFF
--- a/src/cmdstan/command.hpp
+++ b/src/cmdstan/command.hpp
@@ -121,8 +121,11 @@ namespace cmdstan {
       return err_code;
 
     int_argument* random_arg = dynamic_cast<int_argument*>(parser.arg("random")->arg("seed"));
+    unsigned int random_seed;
     if (random_arg->is_default()) {
-      random_arg->set_value((boost::posix_time::microsec_clock::universal_time() - boost::posix_time::ptime(boost::posix_time::min_date_time)).total_milliseconds());
+      random_seed = (boost::posix_time::microsec_clock::universal_time() - boost::posix_time::ptime(boost::posix_time::min_date_time)).total_milliseconds();
+    } else {
+      random_seed = static_cast<unsigned int>(random_arg->value());
     }
     parser.print(info);
     info();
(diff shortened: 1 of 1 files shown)


rok-cesnovar:
@betanalpha @mitzimorris @bob-carpenter @syclik Does anyone have time to review this? This issue   is blocking development on Math and Stan as its causing develop to fail.

PR #733, opened 2019-09-19
Substitute backslashes with slashes for stanc arguments

### LINKED ISSUE #720: Backslash in paths generates C++ error
#### Summary:

I am using cmdStan in Julia in Windows 10. I apologize if this is a Julia related error. What happens is that when Stan generates the file for C++ it uses backslash in the file path, and this makes C++ throw an error.

#### Description:

I am not sure if this is an issue with cmdstan or stan.jl, so if I am opening the issue in the wrong place I apologize.

I am using Julia 1.0.0, cmdStan 2.20.0 and all packages on Julia are up to date. I am using Windows and did the whole work of using RTools to make cmdStan work.

I am using the Bernoulli example that is [in the documentation](https://stanjulia.github.io/Stan.jl/latest/WALKTHROUGH.html). It throws an error and the log says:

```
C:/Users/User/Documents/GitHub/Julia-Para-Economistas/tmp/bernoulli.hpp:21:37: error: incomplete universal character name \U
     reader.add_event(13, 11, "end", "C:\Users\User\Documents\GitHub\Julia-Para-Economistas\tmp\bernoulli.stan");
```

The issue is that for the paths, like `"C:\Users\User\Documents\GitHub\Julia-Para-Economistas\tmp\bernoulli.stan"`, stan is using backslash. Going in the bernoulli.hpp file and changing the backslash in the paths to slash solves the problem. 

So, the question is how do I make cmdStan make this change automatically? Is this an Julia issue or a cmdStan problem?

#### Reproducible Steps:

It is in Julia:

```
using CmdStan
using Mamba
using Cairo

set_cmdstan_home!("C:\\Users\\User\\Downloads\\cmdstan-2.20.0")

bernoullistanmodel = "
data { 
  int<lower=0> N; 
  int<lower=0,upper=1> y[N];
} 
parameters {
  real<lower=0,upper=1> theta;
} 
model {
  theta ~ beta(1,1);
    y ~ bernoulli(theta);
}"

stanmodel = Stanmodel(name="bernoulli", model=bernoullistanmodel)

bernoullidata = Dict("N" => 10, "y" => [0, 1, 0, 1, 0, 0, 0, 0, 0, 1])

rc, sim1 = stan(stanmodel, [bernoullidata], 
"C:\\Users\\User\\Documents\\GitHub\\Julia-Para-Economistas")
```

You can use Stan instead of CmdStan packages, both generate the same error

#### Current Output:

Julia throws _An error has occurred_. The log shows the following error:

```C:/Users/User/Documents/GitHub/Julia-Para-Economistas/tmp/bernoulli.hpp:21:37: error: incomplete universal character name \U
     reader.add_event(13, 11, "end", "C:\Users\User\Documents\GitHub\Julia-Para-Economistas\tmp\bernoulli.stan");
```

#### Expected Output:

Model should be estimated correctly, as it is when I change the paths by hand. 

#### Additional Information:

Maybe I did something wrong when I installed it? It took me a while to get cmdStan to find g++ in RTools so it could compile something. I am not sure if I did something wrong during installation and if it is necessary, I can describe every step I took during instalation. 

#### Current Version:
v2.20.0

Thanks in advance!

rok-cesnovar:
#### Submisison Checklist

#### Summary:

Fixes #720 by replacing \ with / in the path arguments to stanc. It will replace `\\` with `//` but that is not an issue. I am not sure how relevant the issue is given that Stanc3 is really really close and the issue might be solved there, but its still worth fixing I guess. Its better to have this fixed on the interface level anyways I think.

EDIT: Although a fix on the Stan compiler level is a solution for all interface so idk

#### Intended Effect:

Fix the backslash bug for Windows.

#### How to Verify:

Try to compile the bernoulli example on Windows with absolute paths and backslashes or double backslashes in the path to the .stan file.

#### Side Effects:

There should not be any side effects. A minor one is that we try a substitution even on linux and mac where \ or `\\` ar not valid path characters. If this is a problem I can add an if to only do this on Windows.

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company): Rok Češnovar, University of Ljubljana

### DIFF
--- a/make/program
+++ b/make/program
@@ -30,7 +30,7 @@ endif
 %.hpp : %.stan bin/stanc$(EXE)
 	@echo ''
 	@echo '--- Translating Stan model to C++ code ---'
-	$(WINE) bin/stanc$(EXE) $(STANCFLAGS) --o=$@ $<
+	$(WINE) bin/stanc$(EXE) $(STANCFLAGS) --o=$(subst  \,/,$@) $(subst  \,/,$<)
 
 
 .PRECIOUS: %.hpp


seantalts:
Doesn't hurt to fix it here too, as long as the old and new compiler can deal with Linux paths when on Windows. Would you mind testing with the stanc3 stanc dropped in to `bin/stanc` as well? From here: https://github.com/stan-dev/stanc3/releases/tag/v1.0.0-rc3

rok-cesnovar:
I tested it and stanc3 has the same issue and the same fix works so we are good.

I had issues that the stanc exe was overwritten every time so I had to comment that part of makefiles out. But that is a separate issue we will handle later..

seantalts:
Weird! Sounds like the dependencies aren't working on Windows. Very good to
know.

rok-cesnovar:
I will take a look in the evening to see if I can find a simple fix.

rok-cesnovar:
The stanc3 dependency issue is not happening on the stanc3-switch branch so we are fine on that front and this fix works for both.

PR #782, opened 2019-12-18
Replace Eigen vectors of strings with std::vector<std::string>

### LINKED ISSUE #781: clean up uses of Eigen vectors of strings
#### Summary:
Eigen vectors of strings should be replaced by `std::vector<std::string>`.

#### Description:
There are a few uses of `Eigen::Matrix<std::string, Eigen::Dynamic, 1>` in 
- `print.cpp`
- `stansummary.cpp`
- `stansummary_helper.hpp`.

This usage is unconventional and should be replaced with `std::vector<std::string>`.

#### Expected Output:
There should be no changes in output, this is only code cleanup.

#### Current Version:
v2.21.0

mcol:
#### Submisison Checklist

- [X] Run tests: `./runCmdStanTests.py src/test`

#### Summary:
Fixes #781. As I added the `#include <vector>` I also reordered the headers alphabetically.

#### How to Verify:
Existing tests should pass, this is only code cleanup.

#### Side Effects:
None.

#### Documentation:
None.

#### Copyright and Licensing

Marco Colombo

### DIFF
--- a/src/cmdstan/print.cpp
+++ b/src/cmdstan/print.cpp
@@ -1,11 +1,12 @@
 #include <algorithm>
-#include <iostream>
+#include <fstream>
 #include <iomanip>
 #include <ios>
-#include <stan/mcmc/chains.hpp>
+#include <iostream>
+#include <vector>
 #include <cmdstan/print_helper.hpp>
 #include <cmdstan/stansummary_helper.hpp>
-#include <fstream>
+#include <stan/mcmc/chains.hpp>
 
 /**
  * The Stan print function.
@@ -105,8 +106,8 @@ int main(int argc, const char* argv[]) {
 
 
   // Prepare values
-  int n = 9;
-  
+  const int n = 9;
+
   Eigen::MatrixXd values(chains.num_params(), n);
   values.setZero();
   Eigen::VectorXd probs(3);
@@ -127,12 +128,10 @@ int main(int argc, const char* argv[]) {
   }
   
   // Prepare header
-  Eigen::Matrix<std::string, Eigen::Dynamic, 1> headers(n);
-  headers << 
-    "Mean", "MCSE", "StdDev",
-    "5%", "50%", "95%", 
-    "N_Eff", "N_Eff/s", "R_hat";
-  
+  std::vector<std::string> headers = {"Mean", "MCSE", "StdDev",
+                                      "5%", "50%", "95%",
+                                      "N_Eff", "N_Eff/s", "R_hat"};
+
   // Set sig figs
   Eigen::VectorXi column_sig_figs(n);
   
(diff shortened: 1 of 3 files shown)


src/cmdstan/print.cpp
```diff
@@ -127,12 +128,10 @@ int main(int argc, const char* argv[]) {
   }
   
   // Prepare header
-  Eigen::Matrix<std::string, Eigen::Dynamic, 1> headers(n);
-  headers << 
-    "Mean", "MCSE", "StdDev",
-    "5%", "50%", "95%", 
-    "N_Eff", "N_Eff/s", "R_hat";
-  
+  std::vector<std::string> headers = {"Mean", "MCSE", "StdDev",
```
bob-carpenter: [optional]
The `=` is optional here; without it, the construct is called directly.

src/cmdstan/print.cpp
```diff
@@ -105,8 +106,8 @@ int main(int argc, const char* argv[]) {
 
 
   // Prepare values
-  int n = 9;
-  
+  const int n = 9;
```
bob-carpenter: [optional]
If you're setting up constant primitives, they can be declared as static.  It's not enough of a big deal in performance here to be a required change.

src/cmdstan/stansummary.cpp
```diff
@@ -127,12 +128,10 @@ int main(int argc, const char* argv[]) {
   }
   
   // Prepare header
-  Eigen::Matrix<std::string, Eigen::Dynamic, 1> headers(n);
-  headers << 
-    "Mean", "MCSE", "StdDev",
-    "5%", "50%", "95%", 
-    "N_Eff", "N_Eff/s", "R_hat";
-  
+  std::vector<std::string> headers = {"Mean", "MCSE", "StdDev",
```
bob-carpenter: Lists of strings like this shouldn't be duplicated.  But the bigger question is whether these change depending on the quantiles requested.  I wouldn't have thought 5, 50, and 95 could be locked in.

mcol: Most of `print.cpp` is duplicate of `stansummary.cpp`, and is deprecated and already planned to be removed (#153). As such I don't think it's worth spending time to avoid duplication of those strings across the two files. But let me know what you think.

As for the hardcoding of percentiles, there's no way to request other percentiles through the current `stansummary` options (but should be doable via cmdstanr).

src/cmdstan/print.cpp
mcol: Done.

src/cmdstan/print.cpp
mcol: Done.

src/cmdstan/stansummary.cpp
bob-carpenter: OK.  I can just approve if it's already scheduled for demolition.

PR #830, opened 2020-02-21
Add info on threading, mpi and opencl to the csv and stdout

### LINKED ISSUE #793: Revisit printing info on threads and OpenCL devices
#### Description:

In #766 I added some output that would print information on threading and OpenCL. 
Example for both:

```
STAN_THREADS is enabled. map_rect will run with at most 2 thread(s).

STAN_OPENCL is enabled. OpenCL supported functions will use:
Platform: AMD Accelerated Parallel Processing
Device: gfx906
```

In https://discourse.mc-stan.org/t/new-message-on-stan-threads/12596 @wds15 raised some fair question on this solution, mainly the following two:

> I actually think that this statement should be part of the stan services which do printout of the chain run parameters. I mean, the exact numerical results will depend on the choice of the number of threads.

> In case the user turns on threading and MPI, then MPI will be used for parallelism (and only a nested map_rect would use threads). Turning on both is not recommended, but it should work.

While the fix for the latter one can be fixed during the feature freeze, which is currently ongoing, I am not sure we can fix the first one before the next release as that does changes some features of the chain run parameters. So we should probably just revert my PR and handle this again after the release?

#### Current Version:
v2.21.0

rok-cesnovar:
#### Submisison Checklist

#### Summary:

Closes #793 by adding the following info to the stdout and the csv file: 
- number of threads used (STAN_NUM_THREADS if STAN_THREADS is on and STAN_MPI isnt on)
- mpi_enabled = 1 if STAN_MPI is used
- opencl device and platform info if STAN_OPENCL is on

Examples from csv files (it displays the same info in the stdout): 
threads:
```
# output
#   file = output.csv (Default)
#   diagnostic_file =  (Default)
#   refresh = 100 (Default)
# num_threads = 4
```
opencl:
```
# output
#   file = output.csv (Default)
#   diagnostic_file =  (Default)
#   refresh = 100 (Default)
# num_threads = 4
# opencl_platform = AMD Accelerated Parallel Processing
# opencl_platform = gfx906+sram-ecc
```
MPI: 
```
# output
#   file = output.csv (Default)
#   diagnostic_file =  (Default)
#   refresh = 100 (Default)
# mpi_enabled = 1
```

It also removes the stdout info printed in 2.22.

#### Copyright and Licensing

### DIFF
--- a/src/cmdstan/command.hpp
+++ b/src/cmdstan/command.hpp
@@ -8,6 +8,8 @@
 #include <cmdstan/arguments/arg_output.hpp>
 #include <cmdstan/arguments/arg_random.hpp>
 #include <cmdstan/write_model.hpp>
+#include <cmdstan/write_opencl_device.hpp>
+#include <cmdstan/write_parallel_info.hpp>
 #include <cmdstan/write_stan.hpp>
 #include <cmdstan/io/json/json_data.hpp>
 #include <stan/callbacks/interrupt.hpp>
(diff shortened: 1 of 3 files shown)


rok-cesnovar:
@wds15 do you feel better about this way of outputting the threading info?

Issue #861, opened 2020-04-21
makefile variable `STAN_THREADS` on by default


mitzimorris:
#### Summary:

Change makefile so that variable `STAN_THREADS` is always on by default.

#### Description:

Compilation of a Stan model is now independent of compilation of `command.o` and the rest of the stan libs, however, if there is a mismatch between the `STAN_THREADS` compiler flags for the two compilation events, compilation fails with a linker error.

The static member variable instance__ in https://github.com/stan-dev/math/blob/develop/stan/math/rev/core/autodiffstackstorage.hpp is defined differently with one flag or another.

#### Reproducible Steps:

On Linux:  

- clean old builds:  `make clean-all`
- verify that `-DSTAN_THREADS` is undefined:  `make print-compiler-flags`
- build using defaults: `make -j2 build`
- compile model with threading on:  `make STAN_THREADS=T examples/bernoulli/bernoulli`

Fails with message:

```
/usr/bin/ld: _ZN4stan4math22AutodiffStackSingletonINS0_4variENS0_15chainable_allocEE9instance_E: TLS definition in examples/bernoulli/bernoulli.o section .tbss._ZN4stan4math22AutodiffStackSingletonINS0_4variENS0_15chainable_allocEE9instance_E[_ZN4stan4math22AutodiffStackSingletonINS0_4variENS0_15chainable_allocEE9instance_E] mismatches non-TLS reference in src/cmdstan/main.o
src/cmdstan/main.o: error adding symbols: Bad value
collect2: error: ld returned 1 exit status
make/program:48: recipe for target 'examples/bernoulli/bernoulli' failed
make: *** [examples/bernoulli/bernoulli] Error 1
```

#### Current Output:
See above

#### Expected Output:

Setting `STAN_THEADS=TRUE` in CmdStan makefile will override possibility of mismatch.

#### Additional Information:
Provide any additional information here.

#### Current Version:
v2.22

PR #862, opened 2020-04-21
Feature/861 stan threads always on

mitzimorris:
#### Submisison Checklist

#### Summary:

Add logic to CmdStan makefile so that makefile variable `STAN_THREADS` is set to `TRUE`.

#### Intended Effect:

Avoid linker errors.  see issue #861 

#### How to Verify:

Unit tests

#### Side Effects:

Some models may run slower, see discussion here:  https://discourse.mc-stan.org/t/makefiles-and-compiler-flag-dstan-threads-always-on/14482/5?u=mitzimorris

#### Documentation:

NA - compiler flag `STAN_THREADS` no mentioned in user's guide.

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company):  Columbia University
 
### DIFF
--- a/makefile
+++ b/makefile
@@ -18,6 +18,7 @@
 # The default target of this Makefile is...
 help:
 
+
 -include $(HOME)/.config/stan/make.local  # user-defined variables
 -include make/local                       # user-defined variables
 
@@ -30,6 +31,10 @@ endif
 O_STANC ?= 0
 INC_FIRST ?= -I src -I $(STAN)src -I $(RAPIDJSON)
 USER_HEADER ?= $(dir $<)user_header.hpp
+
+ifeq ($(origin STAN_THREADS), undefined)
+STAN_THREADS=TRUE
+endif
 
 
 -include $(MATH)make/compiler_flags


wds15:
Does this qualify for 2.23? Reduce sum will make people want to use this.

makefile
```diff
@@ -31,6 +32,10 @@ O_STANC ?= 0
 INC_FIRST ?= -I src -I $(STAN)src -I $(RAPIDJSON)
 USER_HEADER ?= $(dir $<)user_header.hpp
 
+ifeq ($(origin STAN_THREADS), undefined)
```
rok-cesnovar, line 35: Could this just be `STAN_THREADS ?= TRUE`

rok-cesnovar:
> Does this qualify for 2.23? Reduce sum will make people want to use this.

Given that reduce_sum is the by far biggest new feature of 2.23 I would be for merging this.

makefile
wds15: I would think the same. `STAN_THREADS ?= true` should do it.

wds15:
I just restarted the tests as the performance tests are counter to what I recall.

I am good with making STAN_THREADS on by default; it's just that seeing the performance results makes me cautious.

rok-cesnovar:
I ran longer tests:
Linux - 
https://jenkins.mc-stan.org/job/CmdStan%20Performance%20Tests/job/Custom/153/artifact/linux_compare_results.xml/*view*/

Mac - https://jenkins.mc-stan.org/job/CmdStan%20Performance%20Tests/job/Custom/153/artifact/macos_compare_results.xml/*view*/

Seems worse than we thought it is.

wds15:
Yeah... I am coming to the same conclusion. I wonder if optimisation which we have done in the past releases were more beneficial for non-threading code. Is it possible to see over time (or by release) the performance of individual models? I am just curious.

In any case, I don't think that we can just switch to threads being on for the moment being. In particular, it should not be done as a last minute decision for the release.

Also: I think we should be able to have an even better setup:

- we build the pre-build stuff with and without threads
- if the user does not define STAN_THREADS, then we simply auto-detect it from the model. That is relatively easy for the moment since this is merely a grep for "map_rect" or "reduce_sum". If these are found in the model, then you get threading otherwise not.
- Whenever the user defines STAN_THREADS, then he gets what he asks for

mitzimorris:
> Also: I think we should be able to have an even better setup:
> 
> * we build the pre-build stuff with and without threads
> * if the user does not define STAN_THREADS, then we simply auto-detect it from the model. That is relatively easy for the moment since this is merely a grep for "map_rect" or "reduce_sum". If these are found in the model, then you get threading otherwise not.
> * Whenever the user defines STAN_THREADS, then he gets what he asks for

interesting idea - I'll investigate.

mitzimorris:
just to clarify - let's not merge this PR and I will continue to work on logic.

wds15:
If you want I can give it a try... but give me some days for it.

PR #875, opened 2020-05-18
Feature/872 Precompiled headers for g++

### LINKED ISSUE #872: Precompiled header with g++
#### Description:

We are currently using precompiled model headers (model_header.hpp.gch) only with clang. See lines 20-23 here: https://github.com/stan-dev/cmdstan/blob/develop/make/program#L20

We could do the same with `g++`  which would speedup `g++` compilation of a model to match with clang - see https://github.com/stan-dev/cmdstan/issues/863#issuecomment-622392026.

I tried this out on Linux and Windows. Here is what I found out (I used the example bernoulli model):

- Ubuntu g++ 7.2.0: compile time goes from 12.5s to 6s
- Windows with Rtools 4.0 (g++ 8): compile time goes from 22s to 13s
- Windows with Rtools 3.5 (g++ 4.9.3): does not work

The precompiled header with g++ does take up 500-700MB (the clang++ one takes 170MB on Linux). There is also a warning due to an Eigen issue that needs to be silenced with` -Wno-ignored-attributes` (see comment below).

Given the size of the precompiled header, the warnings that need to be turned off, the fact that it does not work on all of our supported systems and that we can't detect which compiler we use with makefiles, we have the following options:

- A: use it by default on Linux and Mac (do people use g++ on a Mac at all?), and hide it behind a makefile flag (`PRECOMPILED_HEADERS=true` for example) so users on Windows with RTools 4.0 can use it.
- B: hide it behind a makefile flag for g++ on all OSes
- C: not bother with this, close issue

#### Current Version:
v2.23.0

rok-cesnovar:
#### Submisison Checklist

#### Summary:

Fixes #872 by adding precompiled headers for g++.

The implemented logic is as follows:

1. Has the user set the makefile variable `PRECOMPILED_HEADERS`? If yes, use the supplied value and proceed to step 3. If not go to step 2.
2. Is the user on Windows? If yes, set `PRECOMPILED_HEADERS=false`. If not, set `PRECOMPILED_HEADERS=true`
3. If `PRECOMPILED_HEADERS=true` add the precompiled model header as a dependency to the model executable. If `PRECOMPILED_HEADERS` is set to anything else, do not add the dependency.

Additionally if you are using g++ this adds a the "-Wno-ignore-attributes" which silences the g++ warnings seen [here](https://github.com/stan-dev/cmdstan/issues/872#issuecomment-626764948),

The reasoning behind this three step logic is that given that g++ precompiled headers are rather large (500-700MB) we give the users the option to turn them off if they wish to. But the default is to build it if you are not using Windows.

**How to test:**

If you are using clang, just test that your normal workflow still works. Nothing changed there.
If you are using g++ on Linux or Mac then run the following commands on develop and on this branch

```
touch examples/bernoulli/bernoulli.stan
time make examples/bernoulli/bernoulli
```
You should see a 50% decrease in compile time, apart from the first compile where the precompiled header gets built.

If you are using windows you first need to put `PRECOMPILED_HEADERS=true` in the make/local file. If you will not add that, things should work as before.

#### Copyright and Licensing

### DIFF
--- a/make/program
+++ b/make/program
@@ -52,12 +52,11 @@ endif
 
 ifeq ($(PRECOMPILED_HEADERS),true)
 PRECOMPILED_MODEL_HEADER=$(STAN)src/stan/model/model_header.hpp.gch
-CXX
-else
-PRECOMPILED_MODEL_HEADER=
 ifeq ($(CXX_TYPE),gcc)
 CXXFLAGS_PROGRAM+= -Wno-ignored-attributes
 endif
+else
+PRECOMPILED_MODEL_HEADER=
 endif
 
 %.d: %.hpp
--- a/make/program
+++ b/make/program
@@ -17,11 +17,6 @@ $(STAN)src/stan/model/model_header.hpp.gch : $(STAN)src/stan/model/model_header.
 	@echo '--- Compiling pre-compiled header ---	'
 	$(COMPILE.cpp) $< $(OUTPUT_OPTION)
 
-ifeq ($(CXX_TYPE),clang)
-CXXFLAGS_PROGRAM += -include-pch $(STAN)src/stan/model/model_header.hpp.gch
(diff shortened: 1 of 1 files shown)


rok-cesnovar:
This is ready for review.

PR #882, opened 2020-05-30
Using STAN_THREADS, STAN_MPI and STAN_OPENCL without rebuilding

rok-cesnovar:
#### Submisison Checklist

#### Summary:

This PR makes switching on STAN_THREADS, STAN_MPI and STAN_OPENCL much easier.
What it does is changes the dependencies that need to be built for the model exe.

Currently that is:
`main.o` and `model_header.hpp.gch` for everything

Now the the dependency would be:
`main$(STAN_FLAGS).o` and `model_header$(STAN_FLAGS).hpp.gch`

Where `$(STAN_FLAGS)` is built as follows:
```
ifdef STAN_THREADS
STAN_FLAG_THREADS=_threads
endif
ifdef STAN_MPI
STAN_FLAG_MPI=_mpi
endif
ifdef STAN_OPENCL
STAN_FLAG_OPENCL=_opencl
endif

STAN_FLAGS=$(STAN_FLAG_THREADS)$(STAN_FLAG_MPI)$(STAN_FLAG_OPENCL)
```

Without the STAN_THREADS, STAN_MPI or STAN_OPENCL the dependencies are still `main.o` and `model_header.hpp.gch`. 
With STAN_THREADS the dependencies are  `main_threads.o` and `model_header_threads.hpp.gch`.
With MPI its `main_mpi.o` and `model_header_mpi.hpp.gch`. 
With thread and mpi its `main_threads_mpi.o` and `model_header_threads_mpi.hpp.gch`.
With STAN_OPENCL its `main_opencl.o` and `model_header_opencl.hpp.gch`.

The drawback: extra used space, especially the .hpp.gch files. With clang++ each .hpp.gch file is an extra ~200MB, with g++ its more 500-700MB. 

But, for those that dont use flags, nothing changes. And if you dont really use all combinations its not a big deal.

`make clean-all`
`make build -j4`

`time make examples/bernoulli/bernoulli`

Should be fast, the default main.o and precompiled header should have been build by `make build`
```
real    0m7,968s
user    0m7,586s
sys     0m0,349s
```

`touch examples/bernoulli/bernoulli.stan`
`time make STAN_THREADS=true examples/bernoulli/bernoulli`

Will be slower as `main_threads.o` and `model_header_threads.hpp.gch` need to be built.

real    0m43,892s
user    0m42,704s
sys     0m0,923s

`touch examples/bernoulli/bernoulli.stan`
`time make examples/bernoulli/bernoulli`

Should still be fast, nothing changed, main.o and precompiled header should still be there.

real    0m6,211s
user    0m5,988s
sys     0m0,191s

`touch examples/bernoulli/bernoulli.stan`
`time make STAN_THREADS=true examples/bernoulli/bernoulli`

Should be fast, nothing changed, main_threads.o and precompiled header for threads should still be there.

real    0m6,514s
user    0m6,237s
sys     0m0,238s

It also works with STAN_OPENCL and STAN_MPI and any mix of them.

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company): Rok Češnovar, Univ. of Ljubljana

### DIFF
--- a/makefile
+++ b/makefile
@@ -25,6 +25,18 @@ RAPIDJSON ?= lib/rapidjson_1.1.0/
 INC_FIRST ?= -I src -I $(STAN)src -I $(RAPIDJSON)
 USER_HEADER ?= $(dir $<)user_header.hpp
 
+ifdef STAN_THREADS
+STAN_FLAG_THREADS=_threads
+endif
+ifdef STAN_MPI
+STAN_FLAG_MPI=_mpi
+endif
+ifdef STAN_OPENCL
+STAN_FLAG_OPENCL=_opencl
+endif
+
+STAN_FLAGS=$(STAN_FLAG_THREADS)$(STAN_FLAG_MPI)$(STAN_FLAG_OPENCL)
+
 ifeq ($(OS),Windows_NT)
 PRECOMPILED_HEADERS ?= false
 else
@@ -217,8 +229,8 @@ clean-manual:
 
 clean-all: clean clean-deps clean-libraries clean-manual
 	$(RM) bin/stanc$(EXE) bin/stanc2$(EXE) bin/stansummary$(EXE) bin/print$(EXE) bin/diagnose$(EXE)
-	$(RM) -r $(CMDSTAN_MAIN_O) bin/cmdstan
-	$(RM) $(wildcard $(STAN)src/stan/model/model_header.hpp.gch)
+	$(RM) -r src/cmdstan/main.*.o bin/cmdstan
+	$(RM) $(wildcard $(STAN)src/stan/model/model_header.*.hpp.gch)
 	$(RM) examples/bernoulli/bernoulli$(EXE) examples/bernoulli/bernoulli.o examples/bernoulli/bernoulli.d examples/bernoulli/bernoulli.hpp
 
 clean-program:
--- a/makefile
+++ b/makefile
@@ -229,8 +229,8 @@ clean-manual:
 
 clean-all: clean clean-deps clean-libraries clean-manual
 	$(RM) bin/stanc$(EXE) bin/stanc2$(EXE) bin/stansummary$(EXE) bin/print$(EXE) bin/diagnose$(EXE)
-	$(RM) -r src/cmdstan/main.*.o bin/cmdstan
-	$(RM) $(wildcard $(STAN)src/stan/model/model_header.*.hpp.gch)
+	$(RM) -r src/cmdstan/main*.o bin/cmdstan
+	$(RM) $(wildcard $(STAN)src/stan/model/model_header*.hpp.gch)
 	$(RM) examples/bernoulli/bernoulli$(EXE) examples/bernoulli/bernoulli.o examples/bernoulli/bernoulli.d examples/bernoulli/bernoulli.hpp
 
 clean-program:
--- a/makefile
+++ b/makefile
@@ -44,7 +44,7 @@ PRECOMPILED_HEADERS ?= true
 endif
 
 ifeq ($(PRECOMPILED_HEADERS),true)
-PRECOMPILED_MODEL_HEADER=$(STAN)src/stan/model/model_header.hpp.gch
+PRECOMPILED_MODEL_HEADER=$(STAN)src/stan/model/model_header$(STAN_FLAGS).hpp.gch
 ifeq ($(CXX_TYPE),gcc)
 CXXFLAGS_PROGRAM+= -Wno-ignored-attributes
 endif
--- a/.gitignore
+++ b/.gitignore
@@ -11,7 +11,7 @@
 /test/
 
 # main intermediate
-src/cmdstan/main.o
+src/cmdstan/main*.o
 
 # test models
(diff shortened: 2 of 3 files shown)


rok-cesnovar:
This is ready for review. If this is merged, we can close https://github.com/stan-dev/cmdstan/pull/862

mitzimorris:
this is an elegant solution!
we need to figure out where to document the discussion of build times.

rok-cesnovar:
> we need to figure out where to document the discussion of build times.

Not sure what exactly you mean by that. Printed message during compilation or something in the cmdstan manual?. Happy to add more info.

> So if a Stan model includes reduce_sum

I like that idea, though that requires stanc3 passing on some meta info of the model to cmdstan. I have been thinking about that and there are other use cases for that as well, so might be worth a try. For instance a list of data variables, parameters, etc. 

A simple check for inconsistent builds should be added if we will add the threads_per_chaing argument to cmdnstan. But that is a discussion for a separate issue.

mitzimorris:
> Not sure what exactly you mean by that. Printed message during compilation or something in the cmdstan manual?. Happy to add more info.

yes, something in the CmdStan manual (which really needs to be re-worked and put into Rmd - on my to-do list)

referring to discussion in the PR section about calls to make with different combinations of flags - 

> ```
> touch examples/bernoulli/bernoulli.stan
> time make STAN_THREADS=true examples/bernoulli/bernoulli
> ```
>
> Will be slower as main_threads.o and model_header_threads.hpp.gch need to be built.

This could be avoided by making the `build` task build all versions of `main.o` and `model_header.o`

> if a Stan model includes reduce_sum

a quick and dirty approach is to grep the model.hpp looking exactly for string "` reduce_sum(`"  (is that what the call looks like in the generated c++ code?)

rok-cesnovar:
> This could be avoided by making the `build` task build all versions of `main.o` and `model_header.o`

Sure, but I would not want to build too many combinations. I am fine building main.o, main_threads.o and both precompiled headers. Will add that.

mitzimorris:
regarding discussion of does model use `reduce_sum` above, could you please elaborate on: 

> I have been thinking about that and there are other use cases for that as well, so might be worth a try. For instance a list of data variables, parameters, etc.

is this something for the makefile, or something CmdStan should be able to do?  The Stan program exe has methods `get_param_names` `get_constrained_sized_types` (new in stanc3?  how is it used? - stanc2 just had `get_dims`, IIRC)

there's been ongoing discussion around the idea of it would be nice to be able to have an intermediate step between the model compilation step and the estimation step which is just to instantiate the model with the data, in order to validate the inputs.  the problem for CmdStan is that this needs to be done all over again for estimation.

is this the kind of functionality that you're thinking about?

rok-cesnovar:
> I am fine building main.o, main_threads.o and both precompiled headers. Will add that.

On second thought, that is going to be complicated. 
Setting threads and compiling main.o and then unsetting the variable back seems too risky for my blood and not the "makefile" way I guess.

And what if someone calls make build and they have STAN_THREADS in their make/local. With what is currently in this PR, that will build main_threads and precompiled header for threads. And that is fine. But then we have to add logic that would unset the STAN_THREADS variable and also build main.o etc. That all sounds error-prone.

A nice message in the sense of

"First use of Stan makefile flag(s): 'STAN_THREADS' with this installation. Building precompiled headers and object files. This might take a minute."

might suffice and keep us away from the makefile wizardry.

rok-cesnovar:
> (new in stanc3? how is it used? -

Not used anywhere, but could be used for that yes. We can add an additional function that returns other metadata (uses_reduce_sum, uses_map_rect, etc).

mitzimorris:
I think we should keep this PR as simple as possible - yes to adding compiler message when main and header files get compiled; forget about doing more than necessary in `build` task. 
as long as `clean` and `clean-all` tasks get rid of the .o files, all is good.

discussion of how to provide information about compiled model via CmdStan is something we should think about for CmdStan3 et al.  Note that the model only provides functions that reports on output variables - parameters, transformed parameters, generated quantities - not data (input) variables - consider adding these as well?

wds15:
Yeah...we can definitely do more prs which add. This is great as is.

rok-cesnovar:
Added:
`--- Compiling the main object file. This might take up to a minute. ---`
when compiling main*.o
and expanded the pre-compiled header one to:
`--- Compiling pre-compiled header. This might take a few seconds. ---`

makefile
```diff
@@ -25,14 +25,26 @@ RAPIDJSON ?= lib/rapidjson_1.1.0/
...
+STAN_FLAG_THREADS=_threads
+endif
+ifdef STAN_MPI
+STAN_FLAG_MPI=_mpi
+endif
+ifdef STAN_OPENCL
+STAN_FLAG_OPENCL=_opencl
+endif
+
+STAN_FLAGS=$(STAN_FLAG_THREADS)$(STAN_FLAG_MPI)$(STAN_FLAG_OPENCL)
```
mitzimorris, line 38: really nice!

Issue #904, opened 2020-07-07
makefile - ld warnings for bin/stansummary


mitzimorris:
#### Summary:
On MacOS, after changes to  `src/cmdstan/stansummary.cpp`, `make bin/stansummary` results in extensive linker warnings.

command is:
```
clang++ -std=c++1y -Wno-unknown-warning-option -Wno-tautological-compare -Wno-sign-compare -D_REENTRANT       -I stan/lib/stan_math/lib/tbb_2019_U8/include -O3 -I src -I stan/src -I lib/rapidjson_1.1.0/ -I stan/lib/stan_math/ -I stan/lib/stan_math/lib/eigen_3.3.7 -I stan/lib/stan_math/lib/boost_1.72.0 -I stan/lib/stan_math/lib/sundials_5.2.0/include    -DBOOST_DISABLE_ASSERTS            -Wl,-L,"/Users/mitzi/github/stan-dev/cmdstan/stan/lib/stan_math/lib/tbb" -Wl,-rpath,"/Users/mitzi/github/stan-dev/cmdstan/stan/lib/stan_math/lib/tbb"  bin/cmdstan/stansummary.o stan/lib/stan_math/lib/boost_1.72.0/stage/lib/libboost_program_options.a stan/lib/stan_math/lib/boost_1.72.0/stage/lib/libboost_program_options.a         -o bin/stansummary
```
(also note, `stan/lib/stan_math/lib/boost_1.72.0/stage/lib/libboost_program_options.a` listed twice)

output is:
```
[...]

PR #890, opened 2020-06-13
Feature/888 stansummary ci arg

mitzimorris:
#### Submisison Checklist

- [x] Run tests: `./runCmdStanTests.py src/test`
- [x] Declare copyright holder and open-source license: see below

#### Summary:

Added new argument "-percentiles" which takes as value a string consisting
of a set of comma-separated integers between 0 and 100, which will be used
to get the quantiles for all values in the Stan csv output files.

#### Intended Effect:

see #888 

#### How to Verify:

Test via command line:  

 - no --percentiles arg specified, or --percentiles "5,50,95",  - output same as current dev version.

- --percentiles  values "10,50,90",  or "5,20,40,75,99" - etc - will output corresponding quantile values accordingly.

#### Side Effects:

Now using boost::program_options to parse args; makefile changed accordingly.

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company):  Columbia U
By submitting this pull request, the copyright holder is agreeing to license the submitted work under the following licenses:
- Code: BSD 3-clause (https://opensource.org/licenses/BSD-3-Clause)
- Documentation: CC-BY 4.0 (https://creativecommons.org/licenses/by/4.0/)

PR #907, opened 2020-07-11
Feature/903 stansummary sampler diags

mitzimorris:
#### Submisison Checklist

#### Summary:

Modify stansummary console and csv outputs:
 - on the console, there are separate summaries for sampler and model params
 - the csv file contains just the model param summaries

Refactor file `stansummary.cpp`
 - break processing into per-task functions and move from `main` in `stansummary.cpp` into `stansummary_helper.hpp`
 - whenever possible, generalize reporting for both console and csv outputs.

#### Intended Effect:
Improved readability and understandability.
 - summary columns "MCSE", "EFF" and "R-hat" don't make sense for sampler params
 - when calculating column widths and formats, including sampler params in these columns resulted in these values always being reported using scientific notation; omitting this leads to easier to read N_eff and R-hat values.

e.g., current output:
```
~/.cmdstanpy/cmdstan-2.23.0/bin/stansummary output.csv 
Inference for Stan model: lotka_volterra_model
1 chains: each with iter=(1000); warmup=(0); thin=(1); 1000 iterations saved.

Warmup took (6.5) seconds, 6.5 seconds total
Sampling took (6.7) seconds, 6.7 seconds total

                 Mean     MCSE   StdDev     5%    50%    95%    N_Eff  N_Eff/s    R_hat
lp__               28  1.1e-01  2.1e+00     24     29     31  3.3e+02  5.0e+01  1.0e+00
accept_stat__    0.94  2.0e-03  6.7e-02   0.80   0.97   1.00  1.1e+03  1.7e+02  1.0e+00
stepsize__      0.100      nan  1.7e-16  0.100  0.100  0.100      nan      nan      nan
treedepth__       4.8  2.2e-02  6.2e-01    3.0    5.0    5.0  8.2e+02  1.2e+02  1.0e+00
n_leapfrog__       36  5.5e-01  1.6e+01     15     31     63  8.4e+02  1.3e+02  1.0e+00
divergent__      0.00      nan  0.0e+00   0.00   0.00   0.00      nan      nan      nan
energy__          -24  1.6e-01  2.9e+00    -29    -25    -19  3.3e+02  4.9e+01  1.0e+00
theta[1]         0.55  4.8e-03  6.4e-02   0.45   0.55   0.66  1.8e+02  2.6e+01  1.0e+00
theta[2]        0.028  3.0e-04  4.3e-03  0.021  0.028  0.035  2.0e+02  2.9e+01  1.0e+00
theta[3]         0.80  7.0e-03  9.0e-02   0.66   0.79   0.96  1.7e+02  2.5e+01  1.0e+00
theta[4]        0.024  2.7e-04  3.4e-03  0.019  0.024  0.030  1.6e+02  2.4e+01  1.0e+00
```
new output:
```
Input file: output.csv
Inference for Stan model: lotka_volterra_model
1 chains: each with iter=(1000); warmup=(0); thin=(1); 1000 iterations saved.

Warmup took 6.5 seconds
Sampling took 6.7 seconds
                 Mean   StdDev     5%    50%    95%
accept_stat__    0.94  6.7e-02   0.80   0.97   1.00
stepsize__      0.100  1.7e-16  0.100  0.100  0.100
treedepth__       4.8  6.2e-01    3.0    5.0    5.0
n_leapfrog__       36  1.6e+01     15     31     63
divergent__      0.00  0.0e+00   0.00   0.00   0.00
energy__          -24  2.9e+00    -29    -25    -19

                 Mean     MCSE   StdDev     5%    50%    95%  N_Eff  N_Eff/s  R_hat
lp__               28  1.1e-01  2.1e+00     24     29     31    333       50   1.00
theta[1]         0.55  4.8e-03  6.4e-02   0.45   0.55   0.66    176       26    1.0
theta[2]        0.028  3.0e-04  4.3e-03  0.021  0.028  0.035    197       29    1.0
theta[3]         0.80  7.0e-03  9.0e-02   0.66   0.79   0.96    168       25    1.0
theta[4]        0.024  2.7e-04  3.4e-03  0.019  0.024  0.030    162       24   1.00
```

#### How to Verify:
unit tests

#### Side Effects:
console and csv outputs will be different; downstream processing which expects a fixed format may break.

#### Documentation:
online CmdStan User's Guide

#### Copyright and Licensing


mitzimorris:
looking for a reviewer with free cycles this weekend.

it would be lovely to get this in for the code freeze, but I know everyone's super busy and everyone feels this way about their stuff, so if not, that's OK too.

bbbales2:
> the csv file contains just the model param summaries

There's now a csv output of the summary written too? Is there a flag for this? Or was there always this option and I just missed it?

mitzimorris:
> There's now a csv output of the summary written too? Is there a flag for this? Or was there always this option and I just missed it?

yes, this option has been there for a while.  the flag is `--csv_file`.  CmdStanPy uses this to get the summary - perhaps other interfaces do as well.

there's also an option --percentiles="p1,p2,p3,...pN" - this was added recently - https://github.com/stan-dev/cmdstan/pull/890

mitzimorris:
There should be more tests for this feature.  I tested it on a set of 4 csv output files from the Bernoulli model and a single csv output file from the the Lotka-Volterra case study model & data.  the latter is a good test case because it has a parameter which is a 2-D array. the former is also good because a single parameter is the minimal parameters allowed, so it's a good edge case.

bbbales2:
@betanalpha should energy__ be one of the things we compute ESS and Rhat on? Is that useful?

Newline comments:

```
Warmup took 0.0080 seconds
Sampling took 0.012 seconds ########## Put an empty line after this
                Mean   StdDev    5%   50%  95%
accept_stat__   0.91  1.3e-01  0.63  0.97  1.0
stepsize__       1.1  6.7e-16   1.1   1.1  1.1
treedepth__      1.4  4.8e-01   1.0   1.0  2.0
n_leapfrog__     2.4  1.1e+00   1.0   3.0  3.0
divergent__     0.00  0.0e+00  0.00  0.00 0.00
energy__         7.8  1.0e+00   6.8   7.5  9.8

                Mean     MCSE  StdDev     5%   50%   95%  N_Eff  N_Eff/s  R_hat
lp__            -7.3  3.8e-02    0.74   -8.8  -7.0  -6.8    391    32609   1.00
theta           0.25  6.1e-03    0.12  0.076  0.23  0.46    390    32532   1.00
############ Delete an empty line here

Samples were drawn using hmc with nuts.
For each parameter, N_Eff is a crude measure of effective sample size,
and R_hat is the potential scale reduction factor on split chains (at 
convergence, R_hat=1).
```

Calling stansummary with no arguments segfaults:

```
$ bin/stansummary
Segmentation fault (core dumped)
```

Might be the same segfault if the file with the draws doesn't exist:

```
$ bin/stansummary --csv_file csv.csv asdfasdfadsf
csv_filename csv.csv
Input file: asdfasdfadsf
Warning: non-fatal error reading metadata
Error: error reading header
terminate called after throwing an instance of 'std::invalid_argument'
  what():  Error with header of input file in parse
Aborted (core dumped)
```

And I got another segfault trying to write a csv (this is using one chain of the default bernoulli model):
```
$ bin/stansummary --csv_file csv.csv output.csv
csv_filename csv.csv
Input file: output.csv
Inference for Stan model: bernoulli_model
1 chains: each with iter=(1000); warmup=(0); thin=(1); 1000 iterations saved.

Warmup took 0.0080 seconds
Sampling took 0.012 seconds
                Mean   StdDev    5%   50%  95%
accept_stat__   0.91  1.3e-01  0.63  0.97  1.0
...
and R_hat is the potential scale reduction factor on split chains (at 
convergence, R_hat=1).

stansummary: stan/lib/stan_math/lib/eigen_3.3.7/Eigen/src/Core/DenseCoeffsBase.h:118: Eigen::DenseCoeffsBase<type-parameter-0-0, 0>::CoeffReturnType Eigen::DenseCoeffsBase<Eigen::Matrix<double, -1, -1, 0, -1, -1>, 0>::operator()(Eigen::Index, Eigen::Index) const [Derived = Eigen::Matrix<double, -1, -1, 0, -1, -1>, Level = 0]: Assertion `row >= 0 && row < rows() && col >= 0 && col < cols()' failed.
Aborted (core dumped)
```

mitzimorris:
according to Aki: "there is no sense computing or reporting Rhat or ESS for sampler diagnostic columns."
is energy__ not a sampler diagnostic?

https://discourse.mc-stan.org/t/cmdstan-guide-now-online/16368/10

mitzimorris:
@bbbales2 and/or @SteveBronder - fixed newlines, added checks and error messages for all command line args.
ready for code review.

src/cmdstan/stansummary_helper.hpp
```diff
@@ -86,8 +87,7 @@ using Eigen::Dynamic;
 
 Eigen::VectorXi calculate_column_widths(
     const Eigen::MatrixXd &values, const std::vector<std::string> &headers,
-    const int sig_figs,
-    Eigen::Matrix<std::ios_base::fmtflags, Dynamic, 1> &formats) {
+    int sig_figs, Eigen::Matrix<std::ios_base::fmtflags, Dynamic, 1> &formats) {
```
SteveBronder, line 90: Any reason some of these went from const to non-const?

src/cmdstan/stansummary_helper.hpp
```diff
@@ -190,31 +190,434 @@ int matrix_index(std::vector<int> &index, const std::vector<int> &dims) {
...
+std::vector<std::string> get_sampler_params_header(
+    const std::vector<std::string> &percentiles) {
+  // Mean, StdDev, ... percentiles ...
+  std::vector<std::string> sampler_params_header(percentiles.size() + 2);
+  sampler_params_header.at(0) = "Mean";
+  sampler_params_header.at(1) = "StdDev";
+  size_t offset = 2;
+  for (size_t i = 0; i < percentiles.size(); ++i) {
+    sampler_params_header[i + offset] = percentiles[i] + '%';
+  }
```
SteveBronder, line 256:
```suggestion
  for (size_t i = 0; i < percentiles.size(); ++i) {
    sampler_params_header[i + 2] = percentiles[i] + '%';
  }
```

src/cmdstan/stansummary_helper.hpp
```diff
@@ -190,31 +190,434 @@ int matrix_index(std::vector<int> &index, const std::vector<int> &dims) {
...
+  model_params_header.at(1) = "MCSE";
+  model_params_header.at(2) = "StdDev";
+  size_t offset = 3;
+  for (size_t i = 0; i < percentiles.size(); ++i) {
+    model_params_header[i + offset] = percentiles[i] + '%';
+  }
+  offset += percentiles.size();
+  model_params_header.at(offset) = "N_Eff";
+  model_params_header.at(offset + 1) = "N_Eff/s";
+  model_params_header.at(offset + 2) = "R_hat";
```
SteveBronder, line 274:
```suggestion
  for (size_t i = 0; i < percentiles.size(); ++i) {
    model_params_header[i + 3] = percentiles[i] + '%';
  }
  const size_t offset = 3 + percentiles.size();
  model_params_header.at(offset) = "N_Eff";
  model_params_header.at(offset + 1) = "N_Eff/s";
  model_params_header.at(offset + 2) = "R_hat";
```

src/cmdstan/stansummary_helper.hpp
```diff
@@ -190,31 +190,434 @@ int matrix_index(std::vector<int> &index, const std::vector<int> &dims) {
...
+  sampler_params.setZero();
+  for (int i = 0; i < sampler_params.rows(); ++i) {
+    int i_offset = i + sampler_params_start_col;
+    sampler_params(i, 0) = chains.mean(i_offset);
+    sampler_params(i, 1) = chains.sd(i_offset);
+    Eigen::VectorXd quantiles = chains.quantiles(i_offset, probs);
+    for (int j = 0; j < probs.size(); j++)
+      sampler_params(i, 2 + j) = quantiles(j);
+  }
+}
```
SteveBronder, line 291: Am I right that this is checking the cumulative mean, sd, and quantiles of the sampler parameter statistics?

It would be a good sanity check that `sampler_params.size()` is equal to `2 + probs.size()`

src/cmdstan/stansummary_helper.hpp
```diff
@@ -190,31 +190,434 @@ int matrix_index(std::vector<int> &index, const std::vector<int> &dims) {
...
+                          Eigen::MatrixXd &sampler_params) {
+  sampler_params.setZero();
+  for (int i = 0; i < sampler_params.rows(); ++i) {
+    int i_offset = i + sampler_params_start_col;
+    sampler_params(i, 0) = chains.mean(i_offset);
+    sampler_params(i, 1) = chains.sd(i_offset);
+    Eigen::VectorXd quantiles = chains.quantiles(i_offset, probs);
+    for (int j = 0; j < probs.size(); j++)
+      sampler_params(i, 2 + j) = quantiles(j);
+  }
```
SteveBronder, line 290: [optional] Might be easier to read this if the loop is in terms of `quantiles.size()`

src/cmdstan/stansummary_helper.hpp
```diff
@@ -190,31 +190,434 @@ int matrix_index(std::vector<int> &index, const std::vector<int> &dims) {
...
+  Eigen::VectorXd quantiles = chains.quantiles(0, probs);
+  for (int j = 0; j < probs.size(); j++)
+    model_params(0, 3 + j) = quantiles(j);
+  model_params(0, probs.size() + 3) = lp_n_eff;
+  model_params(0, probs.size() + 4) = lp_n_eff / total_sampling_time;
+  model_params(0, probs.size() + 5) = chains.split_potential_scale_reduction(0);
+
+  // Model parameters
+  for (int i = model_params_start_col; i < chains.num_params(); ++i) {
+    int row_offset = i - model_params_start_col + 1;
```
SteveBronder, line 318: I'm having trouble following why the `row_offset` is dictated by the model parameters starting column index.

src/cmdstan/stansummary_helper.hpp
```diff
@@ -50,8 +52,7 @@ int compute_precision(double value, int sig_figs, bool scientific) {
 }
 
 int calculate_column_width(const Eigen::VectorXd &x, const std::string &name,
```
SteveBronder, line 54: Could we give these different names? I just noticed there are `calculate_column_width` and `calculate_column_widths`

src/cmdstan/stansummary_helper.hpp
mitzimorris: the model_params_start column == the number of sampler params + `lp__`

how about I change the loop and define loop variables i,j for the row and column index, respectively.
we start filling in row 1 on param_values, starting at column `model_params_start_col`.

src/cmdstan/stansummary_helper.hpp
mitzimorris: will do.

src/cmdstan/stansummary.cpp
```diff
@@ -53,476 +57,130 @@ int main(int argc, const char *argv[]) {
...
-  }
+  // Validate command line args
   if (vm.count("input_files")) {
+    for (size_t i = 0; i < filenames.size(); ++i) {
+      if (FILE *file = fopen(filenames[i].c_str(), "r")) {
+        fclose(file);
+      } else {
+        std::cout << "Invalid input file: " << filenames[i] << ", exiting." << std::endl;
+        return -1;
+      }
```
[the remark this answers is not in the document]
mitzimorris: used the above based on this advice from SO:  https://stackoverflow.com/questions/12774207/fastest-way-to-check-if-a-file-exist-using-standard-c-c11-c

src/cmdstan/stansummary.cpp
```diff
@@ -53,476 +57,130 @@ int main(int argc, const char *argv[]) {
...
+    } else {
+      std::cout << "Invalid output csv file: " << csv_filename << ", exiting." << std::endl;
+      return -1;
+    }
+    std::cout << "csv_filename " << csv_filename << std::endl;
+  }
+  if (vm.count("sig_figs") && !vm["sig_figs"].defaulted()) {
+    if (sig_figs < 0 || sig_figs > 100) {
+      std::cout << "Invalid --sig_figs option: " << vm["sig_figs"].as<int>()
+                << ", exiting." << std::endl;
```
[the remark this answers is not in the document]
mitzimorris: yes, we should allow 0 significant digits - values greater than 1 or less than -1 are rounded off to whole numbers;
values between -1 and 1 have only 1 non-zero digit.
as for 100 significant digits, this far exceeds numerical accuracy - according to wikipedia, 15-17 significant digits is the max, and for sampling, this requires 10^30 draws.   so really, 5-10 would make a whole lot more sense here.

src/cmdstan/stansummary.cpp
```diff
@@ -53,476 +57,130 @@ int main(int argc, const char *argv[]) {
...
   }
+  if (vm.count("csv_filename")) {
+    if (FILE *file = fopen(csv_filename.c_str(), "w")) {
+        fclose(file);
+    } else {
+      std::cout << "Invalid output csv file: " << csv_filename << ", exiting." << std::endl;
+      return -1;
+    }
+    std::cout << "csv_filename " << csv_filename << std::endl;
+  }
```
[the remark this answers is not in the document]
mitzimorris: same as above.

src/cmdstan/stansummary.cpp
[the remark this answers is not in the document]
mitzimorris: the following line prints out the help msg.  updated description of --sig_figs:
```
 must be in range (0, 10), inclusive
```

src/cmdstan/stansummary.cpp
```diff
@@ -26,503 +28,181 @@ int main(int argc, const char *argv[]) {
...
   if (vm.count("input_files")) {
+    for (size_t i = 0; i < filenames.size(); ++i) {
+      if (FILE *file = fopen(filenames[i].c_str(), "r")) {
+        fclose(file);
+      } else {
+        std::cout << "Invalid input file: " << filenames[i] << ", exiting."
+                  << std::endl;
+        return -1;
+      }
+    }
```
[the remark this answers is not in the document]
mitzimorris: same as above

src/cmdstan/stansummary.cpp
```diff
@@ -26,503 +28,181 @@ int main(int argc, const char *argv[]) {
...
-      std::cout << "input_file ";
+      std::cout << "Input file: ";
     else
-      std::cout << "input_files ";
+      std::cout << "Input files: ";
     size_t i = 0;
-    for (std::vector<std::string>::iterator it = filenames.begin();
-         it != filenames.end(); it++, ++i) {
-      std::cout << (*it);
+    for (size_t i = 0; i < filenames.size(); ++i) {
```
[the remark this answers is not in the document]
mitzimorris: good catch - removed.

src/cmdstan/stansummary.cpp
```diff
@@ -26,503 +26,181 @@ int main(int argc, const char *argv[]) {
...
+    return -1;
+  }
+  if (vm.count("csv_filename")) {
+    if (FILE *file = fopen(csv_filename.c_str(), "w")) {
+      fclose(file);
+    } else {
+      std::cout << "Invalid output csv file: " << csv_filename << ", exiting."
+                << std::endl;
+      return -1;
+    }
```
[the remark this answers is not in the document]
mitzimorris: same reply - I think this is OK.

src/cmdstan/stansummary.cpp
```diff
@@ -26,503 +26,181 @@ int main(int argc, const char *argv[]) {
...
+  }
+  if (vm.count("sig_figs") && !vm["sig_figs"].defaulted()) {
+    if (sig_figs < 0 || sig_figs > 100) {
+      std::cout << "Bad value for option --sig_figs: "
+                << vm["sig_figs"].as<int>() << ", exiting." << std::endl;
+      std::cout << desc << std::endl;
+      return -1;
+    }
+    std::cout << "sig_figs " << vm["sig_figs"].as<int>() << std::endl;
+  }
```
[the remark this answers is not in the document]
mitzimorris: these are all optional arguments - the test `vm.count("arg_name")` is false is the command line has 0 instances of that argument.  if the user should specify `--sig_figs=2` - that's the default, which we know to be acceptable, so no checks required.

src/cmdstan/stansummary.cpp
```diff
@@ -26,503 +26,181 @@ int main(int argc, const char *argv[]) {
...
+                << std::endl;
+      return -1;
+    }
+    std::cout << "csv_filename " << csv_filename << std::endl;
+  }
+  if (vm.count("sig_figs") && !vm["sig_figs"].defaulted()) {
+    if (sig_figs < 0 || sig_figs > 100) {
+      std::cout << "Bad value for option --sig_figs: "
+                << vm["sig_figs"].as<int>() << ", exiting." << std::endl;
+      std::cout << desc << std::endl;
```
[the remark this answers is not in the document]
mitzimorris: the following line prints out the help message, which has been amended, per comment above.

src/cmdstan/stansummary_helper.hpp
mitzimorris: specifying an integer argument value as `const` doesn't mean anything - it's passed by value anyhow.
changed this so that all API calls follow this practice.

src/cmdstan/stansummary_helper.hpp
mitzimorris: implemented suggestion

src/cmdstan/stansummary_helper.hpp
mitzimorris: implemented - omitted `const` - doesn't do anything, code-wise, and not consistent with coding style.

src/cmdstan/stansummary_helper.hpp
mitzimorris: skipping sanity check as this isn't coming from user, it's all internal program logic.

src/cmdstan/stansummary.cpp
```diff
@@ -26,503 +26,181 @@ int main(int argc, const char *argv[]) {
...
+      return -1;
+    }
+    std::cout << "sig_figs " << vm["sig_figs"].as<int>() << std::endl;
+  }
+  if (vm.count("autocorr")) {
+    if (autocorr_idx < 1 || autocorr_idx > filenames.size()) {
```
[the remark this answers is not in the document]
mitzimorris: `autocorr_idx` is the index for the chain.
you're thinking of the lag, which is determined inside the `autocorrelation` function - last function in `stansummary_helper.hpp`.  lines 598-605.   if you have 100-999 draws, lag is 2, 1000 -9999 draws the lag is 3, 10K draws the lag is 4, etc.

this is legacy code, it could be cleaned up, and made a whole lot more general and generally useful.   that's for another issue/PR

src/cmdstan/stansummary_helper.hpp
mitzimorris: you mean quantiles.cols()?  excellent suggestion.  implemented.

src/cmdstan/stansummary_helper.hpp
mitzimorris: cleaned up that function - hope it's clearer.

src/cmdstan/stansummary_helper.hpp
mitzimorris: done.

mitzimorris:
@SteveBronder  - many thanks for slogging through this!  cleaned up a bunch of stuff, per suggestions.  ready for another pass.

mitzimorris:
actually, not ready for review - indexing changes screwed things up.  fixing now.

mitzimorris:
doh!  fixed - ready for re-review.

src/cmdstan/stansummary.cpp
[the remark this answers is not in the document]
SteveBronder: I'd be fine with it being 10 or 20. Dunno why people would ever want/be able to have that many significant digits but maybe someone out there would want that

src/cmdstan/stansummary_helper.hpp
SteveBronder: As long as we are consistent one way or another I'm fine with it

src/cmdstan/stansummary.cpp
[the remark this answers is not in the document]
mitzimorris: in order for 10 significant digits to mean anything, you'd need 10^20 draws from the posterior.
good luck crunching that - I don't think this code base can do it.

mitzimorris:
just added two more unit tests to check that all columns have the right format and content for the console output and the csv_file output. 

these tests run off a saved csv file.  the results of analyzing this file using all default settings should always be the same, given the precision we're reporting.  this should guard against changes to the routines to calculate and summarize the sampler and model params getting the indexing wrong and reporting nonsense.

SteveBronder:
Looking at the changes things look good, but for the tests I'd rather have individual unit tests for the functions instead of regression style tests

bbbales2:
When I did --sig_figs=0 the formatting got weird:

```
                Mean  StdDev  5%  50%  95%
accept_stat__      11.545900e-01   1    1    1
stepsize__         15.330403e-15   1    1    1
treedepth__        14.859980e-01   1    1    2
n_leapfrog__       29.578699e-01   1    3    3
divergent__        00.000000e+00   0    0    0
energy__           89.348156e-01   7    7   10
```

Maybe do `sig_fits` 1 to 10 inclusive?

----

Add a newline after printing errors:
```
bbales2:/home/backup/bbales2/cmdstan-summary$ bin/stansummary --csv_filename
Invalid argument: the required argument for option '--csv_filename' is missingbbales2:/home/backup/bbales2/cmdstan-summary$ 
````

----

--csv_file seems to work for --csv_filename. Can the aliases be listed? Like:

```
--csv_file, --csv_filename       write summary to csv
```

----

Add a colon after `csv_filename`:

```
Input file: output.csv
csv_filename whatever
```

----

By default the sampler csv files are just the last arguments in the list. Should they be documented like that? The docs make it look like you have to include `--input_files` as a flag:

```
 --input_files arg              sampler csv files
```

----

The docs show spaces for percentiles but you can't have spaces when you type the command:

```
--percentiles arg (=5, 50, 95)
```

`--percentiles 5,10,20` and `--percentiles=5,10,20` work.

`--percentiles 5, 10, 20` and `--percentiles=5, 10, 20` do not.

**Comment by mitzimorris** [comment]
> The docs show spaces for percentiles but you can't have spaces when you type the command:
that's a shell parser thing - this will work:  `--percentiles=" 4, 10, 40, 90"`

**Comment by bbbales2** [comment]
> this will work: --percentiles=" 4, 10, 40, 90"

Sure but that's not the example shown

**Comment by mitzimorris** [comment]
> --csv_file seems to work for --csv_filename. Can the aliases be listed?

wow - that's really cool - this seems to be a boost::program_options thing - as long as the prefix is unique, it chooses the right option?  `--csv_f ` also works.  ditto "--c" , so yeah, we could list "--i", "--s", "--a", "--p".

good catch on the inconsistent formatting - fixing now.

**Comment by bbbales2** [comment]
> this seems to be a boost::program_options thing - as long as the prefix is unique

Aaah, okay. I think we should just document the ones we want people to use and not lean on this for anything (since it seems unusual).

**Comment by mitzimorris** [comment]
took a closer look at boost::program_options - short and long options -
how's this?  

```
Allowed options:
  -h [ --help ]                       Produce help message
  -s [ --sig_figs ] arg (=2)          Significant figures reported, (default 
                                      2). Must be in the range (1, 10), 
                                      inclusive.
  -a [ --autocorr ] arg               Display the chain autocorrelation for the
                                      n-th input file, (default none).
  -c [ --csv_filename ] arg           Write model parameter summaries to a csv 
                                      file as well as to console. Will 
                                      overwrite an existing file.
  -p [ --percentiles ] arg (=5,50,95) Percentiles to report, in increasing 
                                      order. Must be integers in the range 
                                      (1,99), inclusive.
  -i [ --input_files ] arg            Sampler csv files. Flag "-i" is optional,
                                      instead list input files at end.

```

**Comment by bbbales2** [comment]
Looking better.

I'm still not a fan of the --input_files thing.

Certainly this stuff isn't standard but usually I know about doing files from a little preamble before the argument docs.

Here's how grep does it on my computer:
```
Usage: grep [OPTION]... PATTERN [FILE]...
Search for PATTERN in each FILE.
Example: grep -i 'hello world' menu.h main.c

Pattern selection and interpretation:
  -E, --extended-regexp     PATTERN is an extended regular expression
  -F, --fixed-strings       PATTERN is a set of newline-separated strings
...
```

Here's emacs:
```
Usage: emacs [OPTION-OR-FILENAME]...

Run Emacs, the extensible, customizable, self-documenting real-time
display editor.  The recommended way to start Emacs for normal editing
is with no options at all.
```

Here's ls:
```
Usage: ls [OPTION]... [FILE]...
List information about the FILEs (the current directory by default).
Sort entries alphabetically if none of -cftuvSUX nor --sort is specified.
```

**Comment by mitzimorris** [comment]
@bbbales2  - is this good enough? 
regarding command line options display - the use of boost::program_options was added in a previous PR in order to allow the user to specify the percentiles (https://github.com/stan-dev/cmdstan/pull/890)
use of the parser was discussed in the issue - (https://github.com/stan-dev/cmdstan/issues/888#issuecomment-642326862) no better solution was available.
this is a *huge* improvement over what came before.
there will be examples in the docs.

**Comment by mitzimorris** [comment]
not really a way to add a preamble to the boost::program_options description object.
but I added more to the input args description, and there will be docs.

```
Allowed options:
  -h [ --help ]                       Produce help message
  -s [ --sig_figs ] arg (=2)          Significant figures reported, (default 
                                      2). Must be in the range (1, 10), 
                                      inclusive.
  -a [ --autocorr ] arg               Display the chain autocorrelation for the
                                      n-th input file, (default none).
  -c [ --csv_filename ] arg           Write model parameter summaries to a csv 
                                      file as well as to console. Will 
                                      overwrite an existing file.
  -p [ --percentiles ] arg (=5,50,95) Percentiles to report, in increasing 
                                      order. Must be integers in the range 
                                      (1,99), inclusive.
  -i [ --input_files ] arg            Sampler csv files. Flag "-i" is optional,
                                      instead list input files at end, e.g.:  
                                      stansummary chain_1.csv chain_2.csv
```

**Comment by mitzimorris** [comment]
I added more more unit tests that step through the process of extracting the info and producing the summary.

**Comment by bbbales2** [comment]
This looks good to me. A couple more small things while we're in this code though.

I'd still like a colon after 'csv_filename' here:

```
$ bin/stansummary --csv afldssadf output.csv
Input file: output.csv
csv_filename afldssadf
Inference for Stan model: bernoulli_model
```

Also it's weird that when I do --sig_figs=0 I get a the help printout:
```
$ bin/stansummary output.csv --sig_figs=0
Input file: output.csv
Bad value for option --sig_figs: 0, exiting.
Allowed options:
  -h [ --help ]                       Produce help message
  -s [ --sig_figs ] arg (=2)          Significant figures reported, (default 
                                      2). Must be in the range (1, 10), 
                                      inclusive.
```

But if I do something else I only get the error (I think only the error is clearer):
```
$ bin/stansummary outasdf.csv
Invalid input file: outasdf.csv, exiting.
```

I'll leave the code review to @SteveBronder since he actually dug into it.

**Comment by bob-carpenter** [comment]
I agree that versions of commands without flags are simpler if they are coherent.  

Consider these two overloads of `tar` and `find` from the unix man pages.

```
tar {-r | -u} -f archive-file [options] [files | directories]

find [-H | -L | -P] [-EXdsx] -f path [path ...] [expression]
```

The `-f` is the "file" argument in both cases.  Both these commands can elide the `-f` using other overloads of the command, which is why everyone's always so confused about order of `tar` arguments (is the output file first or last?).

**Comment by mitzimorris** [comment]
to answer objection that all stats are valid across all params, I've changed the code to compute all stats for all parameters.
putting a space between sampler params and model params.
computing the formats for sampler params and model params separately, as sampler params almost all require use of scientific notation due to nan's in the N_Eff and R_hat columns, which makes it much harder to read these values for model parameters, since they are usually on the scale of 10-10e3.

**Comment by mitzimorris** [comment]
for comparison - output from cmdstan-2.23.0/bin/stansummary on output of eight schools run on example data:
```
Inference for Stan model: eight_schools_model
1 chains: each with iter=(1000); warmup=(0); thin=(1); 1000 iterations saved.

Warmup took (0.056) seconds, 0.056 seconds total
Sampling took (0.046) seconds, 0.046 seconds total

                 Mean     MCSE   StdDev        5%   50%   95%    N_Eff  N_Eff/s    R_hat
lp__              -19  6.7e-01  4.6e+00  -2.6e+01   -19   -11  4.8e+01  1.0e+03  1.0e+00
accept_stat__    0.73  4.8e-02  3.3e-01   3.9e-04  0.89  1.00  4.6e+01  1.0e+03  1.0e+00
stepsize__       0.29      nan  2.8e-16   2.9e-01  0.29  0.29      nan      nan      nan
treedepth__       3.3  7.4e-02  7.3e-01   2.0e+00   3.0   4.0  9.6e+01  2.1e+03  1.0e+00
n_leapfrog__       12  5.0e-01  5.7e+00   3.0e+00    15    15  1.3e+02  2.8e+03  1.0e+00
divergent__     0.021  5.4e-03  1.4e-01   0.0e+00  0.00  0.00  7.1e+02  1.6e+04  1.0e+00
energy__           24  6.7e-01  5.0e+00   1.5e+01    24    32  5.6e+01  1.2e+03  1.0e+00
mu                8.3  2.7e-01  5.1e+00   3.9e-01   8.1    17  3.5e+02  7.7e+03  1.0e+00
theta[1]           12  5.5e-01  8.3e+00   8.0e-01    11    28  2.3e+02  4.9e+03  1.0e+00
theta[2]          8.4  2.9e-01  6.6e+00  -2.4e+00   8.0    19  5.2e+02  1.1e+04  1.0e+00
theta[3]          5.9  3.3e-01  8.5e+00  -9.9e+00   6.3    19  6.5e+02  1.4e+04  1.0e+00
theta[4]          7.9  2.6e-01  6.8e+00  -3.2e+00   7.9    19  6.6e+02  1.4e+04  1.0e+00
theta[5]          4.9  3.2e-01  6.8e+00  -6.9e+00   5.7    15  4.4e+02  9.6e+03  1.0e+00
theta[6]          6.4  3.2e-01  7.2e+00  -5.9e+00   6.3    18  4.9e+02  1.1e+04  1.0e+00
theta[7]           12  4.3e-01  6.9e+00   1.8e+00    11    24  2.6e+02  5.7e+03  1.0e+00
theta[8]          8.9  3.3e-01  7.8e+00  -3.0e+00   8.6    22  5.5e+02  1.2e+04  1.0e+00
tau               7.7  5.1e-01  5.1e+00   1.9e+00   6.4    18  1.0e+02  2.2e+03  1.0e+00

Samples were drawn using hmc with nuts.
For each parameter, N_Eff is a crude measure of effective sample size,
and R_hat is the potential scale reduction factor on split chains (at 
convergence, R_hat=1).
```

output from current version of this PR:

```
Input file: output.csv
Inference for Stan model: eight_schools_model
1 chains: each with iter=(1000); warmup=(0); thin=(1); 1000 iterations saved.

Warmup took 0.056 seconds
Sampling took 0.046 seconds

                 Mean     MCSE   StdDev       5%   50%   95%    N_Eff  N_Eff/s    R_hat

lp__              -19     0.67      4.6      -26   -19   -11       48     1037      1.0
accept_stat__    0.73  4.8e-02  3.3e-01  3.9e-04  0.89  1.00  4.6e+01  1.0e+03  1.0e+00
stepsize__       0.29      nan  2.8e-16  2.9e-01  0.29  0.29      nan      nan      nan
treedepth__       3.3  7.4e-02  7.3e-01  2.0e+00   3.0   4.0  9.6e+01  2.1e+03  1.0e+00
n_leapfrog__       12  5.0e-01  5.7e+00  3.0e+00    15    15  1.3e+02  2.8e+03  1.0e+00
divergent__     0.021  5.4e-03  1.4e-01  0.0e+00  0.00  0.00  7.1e+02  1.6e+04  1.0e+00
energy__           24  6.7e-01  5.0e+00  1.5e+01    24    32  5.6e+01  1.2e+03  1.0e+00

mu                8.3     0.27      5.1     0.39   8.1    17      354     7686     1.00
theta[1]           12     0.55      8.3     0.80    11    28      227     4936     1.00
theta[2]          8.4     0.29      6.6     -2.4   8.0    19      520    11298      1.0
theta[3]          5.9     0.33      8.5     -9.9   6.3    19      647    14056     1.00
theta[4]          7.9     0.26      6.8     -3.2   7.9    19      665    14455     1.00
theta[5]          4.9     0.32      6.8     -6.9   5.7    15      440     9572      1.0
theta[6]          6.4     0.32      7.2     -5.9   6.3    18      493    10716     1.00
theta[7]           12     0.43      6.9      1.8    11    24      260     5652     1.00
theta[8]          8.9     0.33      7.8     -3.0   8.6    22      548    11905      1.0
tau               7.7     0.51      5.1      1.9   6.4    18      103     2235      1.0

Samples were drawn using hmc with nuts.
For each parameter, N_Eff is a crude measure of effective sample size,
and R_hat is the potential scale reduction factor on split chains (at 
convergence, R_hat=1).
```

I think this improves readibility for the model params.
the csv output file remains the same.

**Comment by mitzimorris** [comment]
remaining todos on this PR:
- better usage message
- function-level unit tests for header, stats, and output functions.

**Comment by SteveBronder** [comment]
@mitzimorris ping me when they two above are done and I'll review!

**Comment by mitzimorris** [comment]
Rewrote usage message as follows:
```
> bin/stansummary
Usage: stansummary [OPTIONS] stan_csv_file(s)
Report statistics for one or more Stan csv files from HMC sampler run.
Example:  stansummary model_1.csv model_2.csv
Options:
  -a, --autocorr [n]          Display the chain autocorrelation for the n-th
                              input file, in addition to statistics.
  -c, --csv_filename [file]   Write statistics to a csv file.
  -h, --help                  Produce help message, then exit.
  -p, --percentiles [values]  Percentiles to report as ordered set of
                              comma-separated integers from (1,99), inclusive.
                              Default is 5,50,95.
  -s, --sig_figs [n]          Significant figures reported. Default is 2.
                              Must be an integer from (1, 10), inclusive.

```
@bbbales2 - whaddya think?

**Comment by mitzimorris** [comment]
C++11 raw strings rock!
```
  std::string usage = R"(Usage: stansummary [OPTIONS] stan_csv_file(s)
Report statistics for one or more Stan csv files from HMC sampler run.
Example:  stansummary model_1.csv model_2.csv
Options:
  -a, --autocorr [n]          Display the chain autocorrelation for the n-th
                              input file, in addition to statistics.
  -c, --csv_filename [file]   Write statistics to a csv file.
  -h, --help                  Produce help message, then exit.
  -p, --percentiles [values]  Percentiles to report as ordered set of
                              comma-separated integers from (1,99), inclusive.
                              Default is 5,50,95.
  -s, --sig_figs [n]          Significant figures reported. Default is 2.
                              Must be an integer from (1, 10), inclusive.
)";
  if (argc < 2) {
    std::cout << usage << std::endl;
    return 0;
  }
```

bbbales2:
@mitzimorris very nice!

mitzimorris:
@SteveBronder ready for re-review.

bbbales2:
Running this it looks good.

mitzimorris:
just fixed makefile - stansummary uses boost::program_options, linker was complaining about visibility.
see issue #904

mitzimorris:
@bbbales2 or @SteveBronder - if this builds, ready for re-review.

changed makefile so that changes to `src/cmdstan/stansummary_helper.hpp` will trigger rebuild of `bin/stansummary`
and added compiler flag so that visilbility matches boost::program_options visibility - otherwise ld writes an error novel.

mitzimorris:
not sure what the convention is, but can put the "=" into the docs.

avehtari:
Sorry I was on vacation during the discussion. I make this comment here to ping the people in this thread, but I can also make eventually a separate issue. Would it be possible at some point to change N_Eff to ESS as N is misleading as it in manual refers to to the number iterations in each chain (or is often used for the number of observations). ESS would be more clear as it is acronym for effective sample size. posterior package used .e.g by CmdStanR has at least moved to use ESS.

mitzimorris:
created new issue.

bob-carpenter:
+1 to that.  The problem with "n_eff" is that it's not the number of effective samples.  We consider the set of draws to be a single sample from the posterior.  It's the effective sample size.  

I don't like the `n_` prefix for "number" elsewhere, either, but that's a different issue.

betanalpha:
N_eff and ESS are largely synonymous in practical use and both are regularly 
interpreted as integers.  In particular people read “sample size” and presume 
it is an integer just as strongly as when reading the “N_”.  Moreover when 
the significant figures are set such that the N_eff/ESS column returns integers 
users will often presume an integer meaning regardless.  

The only real solution is to provide better documentation that clarifies what 
each column means, perhaps a —verbose option.

Ultimately I don’t have a problem with changing N_eff to ESS, although it would
break the stansummary UI and user written code that parses for the columns by 
name (same problem for the summary functions in RStan and PyStan).  But it
definitely won’t be sufficient to avoid the “N_eff/ESS is an integer equal to the 
number of independent samples” confusion.

Issue #736, opened 2019-09-20
Make clang++ work on Windows


rok-cesnovar:
#### Summary:

On Windows with `clang++` we get caught in an infinite loop of re-compiling the same files. I exhibited this on clang 9.0.0.

I am investigating it and will try to solve it if I can.

#### Description:

If `CXX=clang++` on Windows `make build` gets caught in an infinite loop of re-compiling all the files. So the same behavior we exhibited on Travis a while ago.

#### Reproducible Steps:

- use Windows
- set CXX=clang++
- make build

#### Current Output:

Infinite loop

#### Expected Output:

Cmdstan should build normally.

#### Current Version:
v2.20.0

bearloga:
I'm also getting this behavior on macOS 10.14, CmdStan 2.20 and clang-7

seantalts:
The infinite loop behavior also shows up when those calls fail in certain
ways. Try commenting out the dependency include from the main makefile and
rerunning make.

seantalts:
@syclik could this be related to the other issues we were seeing before? https://github.com/stan-dev/cmdstan/issues/668 

@bearloga you can try the steps I took in the first post on that PR to see if you have an underlying issue you can fix. especially likely if you recently upgraded Mac OS X or anything like that.

syclik:
> @syclik could this be related to the other issues we were seeing before? #668

Yes, it's possible. I've never run this with clang++ on a Windows box. I'm not good with setup, but if someone can help me log into one, I can do some basic diagnostics and try to see what's causing it. 

If I were debugging, I'd start from a clean build and actually clone the library twice.

In the first one, I'd run `make build` with g++ (or whatever works).

In the second one, I'd run `make build` with clang++ and break after some time.

Then I'd look for the `*.d` files scattered everywhere and do file diffs. I honestly don't know where we'll find differences. It could be something in the stl includes or it could be something else. It has to be there. If it's not in the `*.d` files, that'll tell us something, but I'm not sure what.

Once we find that, we'll need to determine if the dependency files are generated correctly. If they aren't, we can probably tweak the C++ flags or do some `sed` inside the makefile to force it to behave.

Does that info help at all? @rok-cesnovar or @seantalts, do you have boxes that I can log into?

rok-cesnovar:
Thanks @syclik 

I will try your suggestion. I have a Windows box that is currently not on a static IP (in the process of setting that up but university biroucracy is harder to me than integration :) ).

[...]

PR #987, opened 2021-03-06
[WIP] Running multiple chains in one Stan program

SteveBronder:
#### Submisison Checklist

#### Summary:

This is a prototype that can utilize a `chains` and `n_chains` argument to cmdstan so that a single stan program can run multiple chains in parallel like the below

```bash
one_comp sample data file=one_comp.data.R chains n_chains=8
```

#### The Good

It works and gives back correct results! Comparing the results from the multi chain program to single chain programs. This also allows the data in the model to be shared across chains so it should be nice on memory consumption

#### The Bad

It's honestly slower than just running

```bash
seq -w 0 8 | parallel "one_comp sample data file=one_comp.data.R one_comp sample data file=one_comp.data.R"
```

And sometimes by a lot like 20%-30%.

Running this through perf with a few of the models I've added to example files it seems like the slowdown comes from the thread safe function calls like tbb's malloc for threading, `lgamma` and friends in boost, and the other C functions that the  defining the `_REENTRANT` makes thread safe. I'm going to try to profile a few other models and try a few little modifications to see if I can fix those performance losses.

With multiple chains do we also need to think about parallel PRNGs?

My main Q is whether there's enough clever stuff we can do once we have multiple chains in one Stan program to make up for any performance loss. A couple ideas are

1. @wds15 's parallel idea for tree building
2. @bbbales2  and @yizhang-yiz Cross chain warmup 
3. The new variational inference algorithms Akash/Aki/JHuggins and crew are working on (I think this is also a multi-chain thing)
4.  Maybe useful for the adaptation algorithms Liu is working on?
5. Allow OpenCL to share data on the GPU across multiple chains (and generally reduced memory consumption)

#### The Ugly

This is a rough prototype. While the results are correct there's probably some bads happening with the rng. I also wonder about writing to disk across multiple threads. It feels like that could thrash the I/O

#### How to Verify:

You can run the examples I've put in the `examples` folder like

```
# make sure STAN_THREADS=true is defined in make/local
# and 
# TBB_LIBRARIES=tbb tbbmalloc tbbmalloc_proxy
# for tbb_malloc 
export STAN_NUM_THREADS=8

make examples/dawid/dawid
examples/dawid/dawid sample num_samples=150 num_warmup=150 data file=examples/dawid/caries_dump.R chains n_chains=8

make examples/one_comp/one_comp sample
examples/one_comp/one_comp sample num_samples=150 num_warmup=150 data file=examples/one_comp/one_comp.data.R chains n_chains=8

```

Or use perf to check it out like

```
export STAN_NUM_THREADS=8
# If you get a warning about lost chunks turn down freq
perf record -s --per-thread --freq=2800 -g -o test_perf.data examples/dawid/dawid sample num_samples=150 num_warmup=150 data file=examples/dawid/caries_dump.R chains n_chains=8
# -g prints caller graph, -T by thread
perf report -g -T -i ./test_perf.data

perf record -s --per-thread --freq=2800 -g -o test_perf2.data examples/one_comp/one_comp sample num_samples=150 num_warmup=150 data file=examples/one_comp/one_comp.data.R chains n_chains=8

perf report -g -T -i ./test_perf2.data
```

dawid seems to be making a ton of calls to malloc ints. I still need to dive deeper into the perf to figure out where / why that's happening.
 
one_comp is spending most of it's time in `boost::math::lanczos::lanczos_sum` which is called from lgamma

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company): Steve Bronder


SteveBronder:
Also running heap analysis now with

```
STAN_NUM_THREADS=8 heaptrack examples/dawid/dawid sample num_samples=150 num_warmup=150 data file=examples/dawid/caries_dump.R chains n_chains=8
```

and will report back once I have results from both non-parallel and parallel

yizhang-yiz:
> My main Q is whether there's enough clever stuff we can do once we have multiple chains in one Stan program to make up for any performance loss. 

Working on cross-chain warmup tells me that as soon as chains start talking to each other there's lot of information to utilize. Another thing that I came to believe is that massive parallel chains, in the sense that multiple(many more than 4 as in standard practice) communicating chains + within-chain parallelization,  plays a key role to bring large-scale problems I'm interested in to be reasonably within reach.

SteveBronder:
Yeah that's what I was kind of thinking with the part [here](https://github.com/stan-dev/stan/compare/e2cec72bef6e1193368bcb5e6e2c6644d12ecf82...c6491d6e698da669765d892dd8787cf411f61240#diff-eb960ff0bb3205822ea1cd0a58b886985f1497d47d0e723064af6016ef17aad8R124) where we can pretty easily take N warmup samples, do stuff, then run the chains again in an outer for loop around the `parallel_for()`. Once we think we've hit some level of convergence we could even spawn a bunch of new chains based on those points.

There's def a lot of neat things possible with this pattern, but on it's own it's a pretty heavy cost. If we can justify it with some fancy new parallel chain algorithm then I'm 100% for making this look pretty and staring in the deep abyss of perf records to figure out where the bottlenecks are

yizhang-yiz:
> There's def a lot of neat things possible with this pattern, but on it's own it's a pretty heavy cost. 

Agreed. Other than giving us chance to explore, which is important in its own regard, from performance perspective maybe we need think what's the best bang for the buck for a limited # of cores when faced with parallel around chains vs parallel within chains: if a 4-core within-chain parallel run gives speedup 2.0, using that 4 cores for 4 parallel chains better do some really magic on the warmup/sampling.

SteveBronder:
Yes 100%, would you and/or @bbbales2 have time to use this as a base for the `auto` warmup? In my brain I think that's the minimum for this to be reasonable. Though we could also take it a step further and have the user call something like 

```bash
mymodel sample num_warmup=auto num_samples=500 data file='blah.json' chains n_chains=8 max_chains=16
```

Where after the auto warmup we spin off (max_chains - n_chains) additional chains. I think if we can spin those up nicely + doing auto we can get the ESS up to a point that this becomes a very good idea. Going farther we could probably even have something like 

```bash
mymodel sample num_warmup=auto num_samples_ess=300 data file='blah.json' chains n_chains=8 max_chains=16
```

Where we run sampling until the minimum ESS for a parameter is more than `num_samples_ess`

SteveBronder:
Also tagging @LuZhangstat as this may be of interest to her for the adaptation stuff

bbbales2:
> have time to use this as a base for the auto warmup?

Short term no, long term yes. Once we're closer to the other side of some subset of ocl/varmat/tuples/closures/adjoint/laplace.

It's definitely very interesting and an encouragingly small amount of code.

> Where we run sampling until the minimum ESS

That's another idea that to the list.

yizhang-yiz:
> Yes 100%, would you and/or @bbbales2 have time to use this as a base for the auto warmup? 

@SteveBronder I'm booked until May, after that would love to dig into this, hopefully sooner.

LuZhangstat:
> Maybe useful for the adaptation algorithms Lu is working on?
Looks great! Thanks for sharing. I think the implementation of Pathfinder needs GPU parallel computing and clever memory management. Implementing Pathfinder in low-level parallel programming now has low priority in my to do list but I plan to work on it in the near future. By the way, do you happen to know any good tutorial for GPU programming? Many thanks!

SteveBronder:
> Looks great! Thanks for sharing. I think the implementation of Pathfinder needs GPU parallel computing and clever memory management. Implementing Pathfinder in low-level parallel programming now has low priority in my to-do list but I plan to work on it in the near future. By the way, do you happen to know any good tutorial for GPU programming? Many thanks!

Word! If you want GPU stuff I think you can use Stan math's OpenCL backend which has a pretty high level API to do most stuff.

What kinds of operations are you looking to do in particular? If your matrix sizes are not very large it may be better to do something with TBB so you don't pay the transfer cost of moving from the host->GPU

http://mc-stan.org/math/d5/de5/group__opencl.html

@rok-cesnovar we should go through and tag all the GPU functions so they show up there. We should probably do the same thing for the new matrix type then we'd just have an automatic list

For books [OpenCL Programming By Example](http://3.droppdf.com/files/aGywp/opencl-programming-by-example.pdf) is pretty good. The [OpenCL Guide](http://asu-cs.donntu.org/sites/default/files/images/doc/opencl.programming.guide.pdf) is pretty tried and true. I'll try to look around and find what book I used for my GPU course.

LuZhangstat:
@SteveBronder  Thank you so much for sharing the GPU stuff! I will read them later. 

> What kinds of operations are you looking to do in particular? If your matrix sizes are not very large it may be better to do something with TBB so you don't pay the transfer cost of moving from the host->GPU

I will need matrix decomposition like thin QR decomposition and Cholesky decomposition.  The matrix size depends on the number of parameters so it could be large.

SteveBronder:
icic We have the cholesky implemented for gpus in the math library but no thin QR

LuZhangstat:
> We have the cholesky implemented for gpus in the math library but no thin QR

I see, thanks a lot!

SteveBronder:
So, after fixing [#842](https://github.com/stan-dev/stanc3/pull/842) in stanc3 I went back and tried these out again

```bash
# This PR
time -p examples/dawid/dawid sample data file=examples/dawid/caries_dump.R chains n_chains=8
# real 200.60
# user 1452.17
# sys 2.09

# cmdstan develop with gnu parallel
time -p seq -w 0 8 | parallel "examples/dawid/dawid sample data file=examples/dawid/caries_dump.R output file=outputs{}.csv"
# real 246.03
# user 1824.17
# sys 1.52
```

LoL, so that's cool. This went from being "good idea with buts" to just a good idea! I guess it makes since that `current_statement__` constantly changing in the global scope was causing false sharing to happen. This will probably go even faster if I can do [#736](https://github.com/stan-dev/stanc3/issues/736) in stanc3 since that will heavily cut down the programs overall memory footprint.

Anyway, this seems nice but I have no idea how to proceed. My main Qs are around how this should be impl'd in the stan repo. Like should there be a `_par()` equivalent for each of the functions in service? I'm not against writing a design doc for this but would like to have some sort of discussion beforehand.

tagging parties that might be interested @yizhang-yiz  @bbbales2 @wds15

bbbales2:
> current_statement__ constantly changing in the global scope

Wowowow that is a big boost. So now the multithreading is just flat faster than the multiple chains (by a big margin)?

> Anyway, this seems nice but I have no idea how to proceed.

If multithreading is just faster then there is no need to tap into the algorithms above. Otherwise, for justification cross-chain warmup is probably the most ready of those.

As far as coding challenges, I think you're write it'd be a matter of working out how to support the interfaces. Seems like figuring out what potential cmdstan/cmdstanr/cmdstanpy support looks like would be the easiest and then we could work back what would need changed in services from there.

If we went cross-chain warmup route, then we'd need a way to compute cross-chain Rhat and ESS (this code would get used by the variational stuff too).

Yi said he'd be busy until May but it sounds like he'd like to work on it. I like the team projects, so seems fun to me and I'm happy to wait. I think it would be easier to develop on this once varmat is in the bag too since this would be a big thing.

wds15:
This is really awesome work! Given that this shows already benefits - maybe there is a way to get this in quickly in a first version? For example adding an argument `int n_chains = 1` to the current functions so that the default argument will make things just work fine?

Now, there is one thing which I am worried about and that is the nested parallelism pattern which we will end up having when using this with programs based on `reduce_sum` or `map_rect`. The issue is that the TBB runs things *unsequenced*. Have a look here:

https://www.threadingbuildingblocks.org/docs/help/tbb_userguide/work_isolation.html

I do think that we are affected by this and need to be careful as we do use a thread-local ressource (the AD stack). From my understanding we need to ensure that anything within stan-math which fires off sub processes (like a `reduce_sum` call) *must* use work isolation. My suggestion is to wrap reduce calls (and anything else in stan-math which runs parallel things) into an isolated task group as demonstrated in the link above.

Makes sense?

SteveBronder:
> However, in some cases such unsequenced execution may result in errors. For example, a thread-local variable might unexpectedly change its value after a nested parallel construct: 

Ooof yeah if one thread can start writing to another threads stack allocator that could be very bad. But I'm also confused about how that work sharing interacts with thread local storage objects. Will the thread stealing work use the original threads thread local object or will it just not care and use it's own? If the latter then yeah we def need to program against this

 Can we put the `tbb::enumerable_thread_specific<int> ets;` inside of the `ChainableStack` or where would that go? Looking at this it feels like we could just do `tbb::this_task_arena::isolate` inside of `reduce_sum` when we call the parallel reducer. 

> EDIT: An additional point for headaches is MPI... maybe that should just be not allowed with multiple chains. The current MPI design would throw up on this (it would continue working, but only a single chain would ever get to use the workers at any given time).

I've never looked at the MPI stuff but I'll take your word for it. Though how do we program against that so users don't goof? Would it literally just crash or would it just work poorly? I wonder how we can check before we run the chains that the user is not utilizing MPI.

wds15:
The TBB handles **tasks** and abstracts away threads entirely. The only guarantee you have about threads is that they are assigned to task_arena's. No thread can be part of two task_arena's at the same time, but threads can be moved between arenas. One solution is to run different chains in different arenas... but that will not take advantage of resource re-allocation as smoothly, I think (so early finishing chains giving their resources to still running ones).

So this is why we need to isolate any nested parallelism as outlined in the link above. This will give us the right gurantees for the thread local resources. The right action is to do so in reduce_sum and map_rect. At least this is how I understand it.

SteveBronder:
@bbbales2 latest changes should have everything in here. I think since I sorted all this out I'm going to add the dense and other samplers to the Stan PR

bbbales2:
Transformed data runs once per-chain. So it is different than data which is the same for every chain. Since they're both member variables in the model class now, I think one needs to move somewhere else for multi-chain threading to work.

Right now this is used for things like SBC: https://mc-stan.org/docs/2_26/stan-users-guide/sbc-in-stan.html (each chain generates data then does inference on it -- every chain runs its own transformed data block).

SteveBronder:
Yeah @betanalpha also brought this up. The problem is then we need multiple models which I think would remove a lot of the speedup benefit.

But that's also a v big problem, we don't want folks thinking multi chain works for sbc when it won't. The only solution I can think of is

1. Add a function the the model class that, if an _rng function is used in transformed data returns true, otherwise false

2. The first time we construct a model we look at the return of that function and if it's true then we make N models

I think we have to have N models in that case with N rng passed to the model otherwise we'd have to introduce state into the model which would make a whole other set of problems to sort

mitzimorris:
> But that's also a v big problem, we don't want folks thinking multi chain works for sbc when it won't. The only solution I can think of is
> 
> 1. Add a function the the model class that, if an _rng function is used in transformed data returns true, otherwise false
> 2. The first time we construct a model we look at the return of that function and if it's true then we make N models
> 
> I think we have to have N models in that case with N rng passed to the model otherwise we'd have to introduce state into the model which would make a whole other set of problems to sort

at the point you construct N models, you have N objects, each with its own copy of the data.

adding that function to the generated code is, in theory, OK, but it adds complexity and clutter.

so yeah, v big problem.

bbbales2:
> a whole other set of problems to sort

A separate data class would solve this I think. Like if the stanc3 compiler generated:

```
class model_data {
  int N;
  std::vector<double> y;
  // ... all the data block stuff here
};

class model {
  const model_data& model_data_;
  // ... all the transformed data stuff goes here
};
```

So there could be multiple model objects referencing the same underlying data.

And then every time data is referenced in the model, instead of `y` it is `model_data_.y`.

SteveBronder:
> A separate data class would solve this I think. Like if the stanc3 compiler generated:

I feel like that would be kind of hard because then we have to have construct a pointer of type `model_data` in the model block.

I figured out a way in stanc to just check if the model uses any RNG functions in transformed data. We can just make the first model, ask if it uses any RNG functions in transformed data, and then if so make a model for each chain we'll be running

https://github.com/stan-dev/stanc3/pull/868

mitzimorris:
> I figured out a way in stanc to just check if the model uses any RNG functions in transformed data. We can just make the first model, ask if it uses any RNG functions in transformed data, and then if so make a model for each chain we'll be running

serious technical debt here.  the compiler is now just that much more complicated.  if this feature delivers a huge win to some important class of users, great.  if not, is is worth it?

wds15:
We saw 20% or so runtime improvements. That’s a lot for not changing any code of Stan model! So we want this. How about we just document that there will only be one transformed data within any given process? So if chains are run using threads instead of multiple processes then the transformed data will be the same.

most models do not use rng calls in transformed data is what I would guess is the case. So the majority of models would gain a lot here.

bbbales2:
> ask if it uses any RNG functions in transformed data, and then if so make a model for each chain we'll be running

I do not like this. There would inevitably be weird gotchas here. Like, profiling would work differently depending on if you have _rngs in transformed data or not.

If a model is a per-chain thing -- that's fine. That'll actually make threading more like our current thing, and this will be useful as an infrastructure project even if we don't get the shared data benefits immediately. We can come back later and do a shared data object.

> would be kind of hard

Ooo, I didn't mean to imply any tricks. I think it would be quite simple (ignoring compatibility difficulties). Like the code for command.hpp would look like:

```cpp
int main(int argc, char** argv) {
  model_data shared_model_data(data_file);

  std::vector<model> models;
  for(size_t i = 0; i < chains; ++i) {
    models.push_back(model(shared_model_data));
  }

  run_sampler(models);

  // shared_model_data and models all get destructed here
}
```

SteveBronder:
> I do not like this. There would inevitably be weird gotchas here. Like, profiling would work differently depending on if you have _rngs in transformed data or not.

Why would profiling work different? I already have a branch up in stanc that can tell us if an rng was used in transform data.

> Ooo, I didn't mean to imply any tricks. I think it would be quite simple (ignoring compatibility difficulties). Like the code for command.hpp would look like:

Thee use case we are talking about is when the data for the model is made using SBC in transformed data. In that case the data the user gives is probably very small so I'm not really understanding how the above solves our problem?

This also feels like it would be very tedious to do in the compiler. Every time we access data it needs to be through a smart pointer, but transform data stays the same. So every time you use any object in the model block you have to ask if it's data, then if it's data data or transformed data, and then you can print out whether to access it via mydata-> or just the variable name

mitzimorris:
the SBC use case is not at all the same as the others - the regular use case is that you're running a bunch of chains because you want to combine the draws - it's all the same sample.  SBC is something different.

overall, I don't see that the pieces of the puzzle fit together cleanly, which is problematic both for the code and for the users.  we end up with something that's difficult to explain, document, use, and maintain.

bbbales2:
> Why would profiling work different?

If it's one model, all chains accumulate to one profile. If it's separate models, chains accumulate to separate profiles.

> I'm not really understanding how the above solves our problem

If the goal here is to share data between the different chains, and the original plan of sharing the model between all the chains broke because the model itself needs to be a per-chain thing because of transformed data, then we can specifically share just the data between all the chains by sharing the data between all the models.

> This also feels like it would be very tedious to do in the compiler. Every time we access data it needs to be through a smart pointer

I don't know how tedious it would be at the stanc3 compiler level. There would need to be something marking the difference between variables defined in the data block and variables defined in the transformed data block. This already exists for different blocks -- like you can't assign transformed parameters in the model block. Yeah it's work but we also don't have to do it immediately even.

As far as a smart pointer, yes this could be done with a smart pointer, or a pointer, or a const reference. I assume pick whichever is easiest for the C++ compiler to work with.

> I already have a branch up in stanc that can tell us if an rng was used in transform data.

If the goal is to get threads in fast, let's just do separate models completely and then do a shared data optimization later.

rok-cesnovar:
> If the goal is to get threads in fast, let's just do separate models completely and then do a shared data optimization later.

With separate models we probably dont get the speedup?

bbbales2:
> With separate models we probably don't get the speedup?

Yeah, but we can still sort out all the interface stuff and make sure things are threadsafe and whatnot.

wds15:
Are we sure that with separate models there is no speed gain?

is there speed gain with shared data, but chain specific transformed data?

bbbales2:
> is there speed gain with shared data, but chain specific transformed data?

Oh oh, I assume there would also be speedup if we did a shared data object.

Specifically what I meant in [this](https://github.com/stan-dev/cmdstan/pull/987#issuecomment-812903409) response was: the quick implementation is to do separate models each with their own data. The upside to this is working out the interfaces and thread safety stuff. The downside is that we would not expect to have the threading speedup (because I am assuming the original measured speedup has to do with the data being shared).

SteveBronder:
> If it's one model, all chains accumulate to one profile. If it's separate models, chains accumulate to separate profiles.

Can you show me where that's happens? I'm pretty certain having N models will not change anything wrt profilling.

> If the goal here is to share data between the different chains, and the original plan of sharing the model between all the chains broke because the model itself needs to be a per-chain thing because of transformed data, then we can specifically share just the data between all the chains by sharing the data between all the models.

> overall, I don't see that the pieces of the puzzle fit together cleanly, which is problematic both for the code and for the users. we end up with something that's difficult to explain, document, use, and maintain.

I'm confused about why everyone is confused, idt I clarified the problem and what we want to do

Problem:
  If someone uses a random number generator in transformed data, such as for SBC, then we would like to have each chain use a different RNG, and thus a different model.

we want multiple models **only when people use an _rng() in the transformed data block**. If we don't have an _rng() used in the transformed data block then we don't need multiple models and get the benefits of sharing data across the threads.

Doing this with the stanc PR I've posted [here](https://github.com/stan-dev/stanc3/pull/868) in cmdstan or other interfaces is very simple.


```cpp
  // Instantiate model
  std::shared_ptr<stan::model::model_base> &model
      = std::make_shared_ptr<stan::model::model_base>(new_model(*var_context, 
           random_seed, &std::cout));

std::vector<std::shared_ptr<stan::model::model_base>> models;
models.push_back(model);

// Ask if it uses an rng in transformed data
if (*model.is_rng_in_transformed_data()) {
  // true, we make a new model for each chain
  for (int i = 1; i < n_chains; ++i) {
    models.emplace_back(new_model(*var_context, random_seed, &std::cout));
  }
} else {
  // false, we use the same model for all the chains
  for (int i = 1; i < n_chains; ++i) {
    models.push_back(model);
  }  
}
```

That's pretty low overhead and we get the benefits of having one model when we are not doing SBC and then can still do sbc safely.

Is there something that's very bad with this approach? It seems to solve the problem in a simple way and takes about a 100 lines of code in the ocaml compiler

bbbales2:
> Can you show me where that's happens? I'm pretty certain having N models will not change anything wrt profilling.

Whoops, I thought the profile was a member variable in the model class. I'm wrong there.

> I'm confused about why everyone is confused

I don't think I'm confused -- I disagree with the _rng solution.

> we want multiple models only when people use an _rng() in the transformed data block

I disagree that this is the basic problem. The basic problem is adding in threading (the upside coming from the performance gain from shared data and the fact that it serves as infrastructure for other algorithms that can take advantage of cross-chain checks).

> Is there something that's very bad with this approach?

Yes. The problem to solve is how to share the data between multiple chains in a threaded multi-chain Stan. This solution is doing that by detecting when it is possible to do an optimization to share the model itself between multiple threads (or not).

This seems like an really indirect way to solve the problem of sharing model data, when we could just directly address the Stan memory model. Even if the _rng solution makes sense now, it probably won't in a few weeks, or to whomever comes along in the future.

SteveBronder:
> >we want multiple models only when people use an _rng() in the transformed data block

> I disagree that this is the basic problem. The basic problem is adding in threading (the upside coming from the performance gain from shared data and the fact that it serves as infrastructure for other algorithms that can take advantage of cross-chain checks).

Oh well I think my own confusion makes more sense now because we were giving answers two totally different questions lol. 

I'd push back on threading being the problem we are trying to solve here with any talk about data. Like this branch 100% works and does stuff computationally correct and statistically correct, but looking at the [Stan docs for block actions](https://mc-stan.org/docs/2_26/reference-manual/overview-of-stans-program-blocks.html) `data` and `transformed data` are supposed to happen once per chain and we are talking about how to handle that. The easy way is to make N models for each N thread, but we also see that the only time we need to actually do that is when a user is using an `_rng()` in the transformed data block (or is that not correct and there are other cases where data / transformed data would differ across chains?). At this point it feels like the question we are trying to answer is what you say below, "In a multi-threaded environment how do we allow data to be shared across threads."

>>    Is there something that's very bad with this approach?

> Yes. The problem to solve is how to share the data between multiple chains in a threaded multi-chain Stan. This solution is doing that by detecting when it is possible to do an optimization to share the model itself between multiple threads (or not).

> This seems like an really indirect way to solve the problem of sharing model data, when we could just directly address the Stan memory model. Even if the _rng solution makes sense now, it probably won't in a few weeks, or to whoever comes along in the future.

What do you mean by indirect? If there's some more obvious way to do this then I'm all for it, but I think it's the only literal way to do the problem of multiple transformed data for SBC besides making N copies of the model. Like, let's say we did the whole thing in stanc to have the model class take in a `model_data` class that actually holds the data. Now a user is going to generate data in transformed data with our `_rng()` functions for SBC. Whether or not we have that transformed data in the model class or the model data class, how are we going to know that we need to replicate transformed data N times because the user used an `_rng()` function in transformed data vs. when can we do the optimization to share the data across threads?

If we do make N copies of the model then honestly idt any of this effort is going to lead to performance gains and then this version of simple parallelism seems like it gives no real benefit. The main savings from this are in reduced memory consumption, especially for v big models like [this](https://discourse.mc-stan.org/t/issue-with-dual-averaging/5995/48?u=stevebronder) which have like 2 million rows of data. Cutting down memory consumption will give very nice performance gains for these large data models.

And like in the meta, I think changing our data memory model is going to be a v big effort. It feels like that takes doing parallism from a few weeks problem to a few months problem. Are there other benefits we would get from separating out the model data from the model class?

nhuurre:
> looking at the Stan docs for block actions data and transformed data are supposed to happen once per chain

That page is inconsistent about that. The block actions table indeed says it's per chain but the **Transfomed variables** section earlier says 
> the transformed data block are only ever evaluated once.

and has even a footnote which is very clear
> If the C++ code is configured for concurrent threads, the data and transformed data blocks can be executed once and reused for multiple chains.

(I don't understand what the footnote is talking about--it sure sounds like this it's very PR but it's been there for years...)

Currently CmdStan has two input arguments that control the rng: `random seed` and chain `id` _but the chain id has no effect on transformed data_.

I think it makes most sense that by default all chains target the same posterior distribution (i.e. share all (transformed) data, even if it came from an `_rng`.)

SteveBronder:
> (I don't understand what the footnote is talking about--it sure sounds like this it's very PR but it's been there for years...)

Yeah the docs are kind of inconsistent around this, but I think it's because it was not thinking about running multiple chains in one program at that time.

> Currently CmdStan has two input arguments that control the rng: random seed and chain id but the chain id has no effect on transformed data.

> I think it makes most sense that by default all chains target the same posterior distribution (i.e. share all (transformed) data, even if it came from an _rng.)

In the meta here if someone does multi-chain with SBC in a higher level package like cmdstanr they would expect the current behavior where N models are made and transformed data is done N times. If we don't account for this behavior in the impl here than I think that things like cmdstanr would have to account for this behavior themselves. Like if cmdstanr wants to do parallel chains but have a new transform data per thread made then they have to go back to running multiple cmdstan processes. 

If there are other edge cases where we need to run td multiple times I'm open to talking about it, but it sounds like it's just the rng() case and I have the stuff already written to detect that which we can plug in here.

mitzimorris:
> Yeah the docs are kind of inconsistent around this, but I think it's because it was not thinking about running multiple chains in one program at that time.

I'm not sure about that - I told Bob about this effort and he said that everything was designed for threading.

mitzimorris:
at this point we really need a design proposal for this feature - I'm trying to follow the different PRs in the different repos. 
 the existance of multiple PRs in multiple repos says that this is a feature that's involved enough to warrent upfront design and documentation, not to mention investigation when this will speed things up, by how much, and when it will hurt.

also, while this feature will speed things up, the CmdStanX interfaces, actually all the interfaces, already allow a user to run chains in parallel, so it seems to me the only users that will really benefit here are folks using CmdStan.   if this feature goes in, it will take work across all the interfaces to expose the potential speedups multithreading offers.

if it's the case that multithreading can potentially slow things down, we'll need to provide guidelines for users. and very few users have the requisite understanding of threads to even understand their options.   in which case,  this is a "power user" feature, which is also fine, but I feel like if there's a feature that only benefits power users, it should do a whole lot more than give them a modest 20% speedup under certain circumstances.

I realize a lot of work has gone into this already.  this has the feel of a feature that is proving to be more complicated than anticipated which is spinning into something big and messy.  could we please step back and assess?

SteveBronder:
> at this point we really need a design proposal for this feature - I'm trying to follow the different PRs in the different repos.
the existance of multiple PRs in multiple repos says that this is a feature that's involved enough to warrent upfront design and documentation, not to mention investigation when this will speed things up, by how much, and when it will hurt.

It's not clear to me what a design doc would do to help the confusion here. The actual implementation is very simple, its mostly now sorting out the odds and ends of edge cases and I'm not sure if a design doc is going to help as we only really catch these things as we code and test the implementation.

imo there's only a few bits remaining to get this to work

1. The names of the arguments for cmdstan (which I think we've agreed on at this point.)
2. Add [this PR](https://github.com/stan-dev/stanc3/pull/868) to stanc so we can detect if we need to make mutliple models
3. Benchmark a model with multi-chain uses reduce sum
4. Merge in the services API for multichain [here](https://github.com/stan-dev/stan/pull/3033)
5. Have cmdstan use the multi chain services (this PR)

> also, while this feature will speed things up, the CmdStanX interfaces, actually all the interfaces, already allow a user to run chains in parallel, so it seems to me the only users that will really benefit here are folks using CmdStan. if this feature goes in, it will take work across all the interfaces to expose the potential speedups multithreading offers.

There is a very very big difference between how we do parallism now and the scheme here in that we can share data across chains. That can be very important for models with a large amount of data as we can now fit multiple chains on systems that we would have previously run out of memory on. Combining multi chain with schemes like [here](https://github.com/stan-dev/stanc3/pull/865) for not having to copy that data onto each thread's stack allocator can make for dramatic reductions in memory usage. Less memory means we can fit larger models and more chains on smaller computers and fetch from higher level memory less often as the computer knows it can use the same memory without having to fetch from higher levels to get the same thing.

Also the PR in (4) is in the services part of Stan, so rstan, httpstan, and cmdstan can all use that service layer for multiple chains in a Stan program.

> if it's the case that multithreading can potentially slow things down, we'll need to provide guidelines for users. 

@wds15 do you have benchmarks laying around for models where turning threading on will lead to performance losses? What functions will that happen to? From what I remember it was only windows? Maybe new versions of mingw fixed those issues.

> if it's the case that multithreading can potentially slow things down, we'll need to provide guidelines for users. and very few users have the requisite understanding of threads to even understand their options. in which case, this is a "power user" feature, which is also fine, but I feel like if there's a feature that only benefits power users, 

I think we will end up making a hard and fast choice to either turn threading on at all times or just simply not do this feature. From what I remember it was only windows users and only the lgamma function that went slower?

>it should do a whole lot more than give them a modest 20% speedup under certain circumstances.

The speedup is fine but its really the huge memory reduction that is the thing here. Like right now if you have a model whose data comes out to be equivalent to like a 2 million row by 60 column matrix that's going to take up about 1 GB of memory. If we run 8 chains that's going to require 8 GB, but with this scheme it's just going to be 1 GB. That's an 8 fold reduction in the amount of memory a stan program needs for multi-chain and I think that's pretty important.

> I realize a lot of work has gone into this already. this has the feel of a feature that is proving to be more complicated than anticipated which is spinning into something big and messy. could we please step back and assess?

I still think the implementation is simple here and we are mostly sorting out edge cases to figure out a few weird behaviors. I think there's been some confusion about a few things that has made this seem more complicated than it actually is.

nhuurre:
When I ask CmdStanPy for multiple chains it gives me ones with identical transformed data
```python
>>>> with open("tdata.stan","w") as f:
....   f.write("""
.... transformed data { real z_ = normal_rng(0,1); }
.... parameters { real<lower=0,upper=1> x; }
.... transformed parameters { real z = z_; }
.... """)
.... 
129
>>>> m=cmdstanpy.CmdStanModel(stan_file="tdata.stan")
INFO:cmdstanpy:compiling stan program, exe file: /home/nhuurre/tdata
INFO:cmdstanpy:compiler options: stanc_options=None, cpp_options=None
INFO:cmdstanpy:compiled model file: /home/nhuurre/tdata
>>>> d=m.sample(chains=4)
INFO:cmdstanpy:start chain 1
INFO:cmdstanpy:start chain 2
INFO:cmdstanpy:start chain 3
INFO:cmdstanpy:start chain 4
INFO:cmdstanpy:finish chain 1
INFO:cmdstanpy:finish chain 2
INFO:cmdstanpy:finish chain 3
INFO:cmdstanpy:finish chain 4
>>>> d.draws()[1,:,d.column_names.index('z')]
array([0.292274, 0.292274, 0.292274, 0.292274])
```
Surely this is the behaviour users expect.
You can still do SBC, you just need to start multiple runs with different random seeds.

SBC could probably benefit a lot from automated warmup and my understanding is that the proposal for _that_ requires cross-chain communication. It only makes sense if you have multiple threads with the _same_ (rng generated) transformed data.

wds15:
Now I recall... the `transformed data` will always be the same whenever the random seed is fixed *for any chain-id*. This is intended by design based on the rationale that "one model, one data". Here is the bernoulli exampe modified:

```stan
data {
  int<lower=0> N;
  int<lower=0,upper=1> y[N];
}
transformed data {
   real lucky = normal_rng(0,1);
   print("My lucky number: ", lucky);
}
parameters {
  real<lower=0,upper=1> theta;
}
model {
  theta ~ beta(1,1);  // uniform prior on interval 0,1
  y ~ bernoulli(theta);
}
```

and here is the output for 2 chains with the same random seed:

```
[09:41:20][sebi@sebastians-macbook-pro:~/work/cmdstan/examples/bernoulli]$ ./bernoulli-test sample data file=bernoulli.data.R output refresh=0 id=0 random seed=1213
...
random
  seed = 1213
output
  file = output.csv (Default)
  diagnostic_file =  (Default)
  refresh = 0
  sig_figs = -1 (Default)
  profile_file = profile.csv (Default)

My lucky number: -2.20795

Gradient evaluation took 9e-06 seconds
1000 transitions using 10 leapfrog steps per transition would take 0.09 seconds.
Adjust your expectations accordingly!



 Elapsed Time: 0.005 seconds (Warm-up)
               0.014 seconds (Sampling)
               0.019 seconds (Total)

[09:42:00][sebi@sebastians-macbook-pro:~/work/cmdstan/examples/bernoulli]$ ./bernoulli-test sample data file=bernoulli.data.R output refresh=0 id=1 random seed=1213
method = sample (Default)
...
random
  seed = 1213
output
  file = output.csv (Default)
  diagnostic_file =  (Default)
  refresh = 0
  sig_figs = -1 (Default)
  profile_file = profile.csv (Default)

My lucky number: -2.20795

Gradient evaluation took 1e-05 seconds
1000 transitions using 10 leapfrog steps per transition would take 0.1 seconds.
Adjust your expectations accordingly!



 Elapsed Time: 0.005 seconds (Warm-up)
               0.015 seconds (Sampling)
               0.02 seconds (Total)
```

So we do not have any problem at all as it looks with transformed data. By design the random seed fixes the data (including the transformed data).

wds15:
> @wds15 do you have benchmarks laying around for models where turning threading on will lead to performance losses? What functions will that happen to? From what I remember it was only windows? Maybe new versions of mingw fixed those issues.

Using a model with `lgamma` will be sensitive to using threading or not to some degree. So just generate one with brms, for example.

Apart from `lgamma` the thread local AD tape does cost some performance. As I recall, the cost is minimal on Linux, very small on MacOS and quite large on Windows. The Windows slowdowns were observed with the old mingw toolchain prior to R 4.0 such that these hopefully come out better with the newer toolchain.

If we can settle on some average performance gain for the threaded setup of running chains, then it would be ok to loose in return a bit of performance for the thread only approach.

(BTW, for rstan we already live in a threading only world as it was too difficult to get things to work otherwise if I recall correctly)

SteveBronder:
> SBC could probably benefit a lot from automated warmup and my understanding is that the proposal for that requires cross-chain communication. It only makes sense if you have multiple threads with the same (rng generated) transformed data.

> So we do not have any problem at all as it looks with transformed data. By design the random seed fixes the data (including the transformed data).

Oh that's awesome cool cool I'm going to close that stanc PR then, thank you both for looking into that!

> (BTW, for rstan we already live in a threading only world as it was too difficult to get things to work otherwise if I recall correctly)

If it's a windows only slowdown and most windows users probably use rstan, I wonder if we should just turn on threading? Because I also don't have access to a windows computer so idk how to do that benchmark.

bbbales2:
> When I ask CmdStanPy for multiple chains it gives me ones with identical transformed data
> Now I recall... the transformed data will always be the same whenever the random seed is fixed for any chain-id. This is intended by design based on the rationale that "one model, one data".

Lolol, ezmoney.

Thanks @nhuurre and @wds15 for investigating.

That solves that easily enough.

mitzimorris:
now that we don't have a problem, I don't have an objection - this all makes sense and is consistent with what Bob says.

mitzimorris:
> imo there's only a few bits remaining to get this to work
> 
> 1. The names of the arguments for cmdstan (which I think we've agreed on at this point.)
> 2. ~~Add [this PR](https://github.com/stan-dev/stanc3/pull/868) to stanc so we can detect if we need to make mutliple models~~
> 3. Benchmark a model with multi-chain uses reduce sum
> 4. Merge in the services API for multichain [here](https://github.com/stan-dev/stan/pull/3033)
> 5. Have cmdstan use the multi chain services (this PR)
> 
> >

striking item 2 from the list, correct?

mitzimorris:
> Currently CmdStan has two input arguments that control the rng: `random seed` and chain `id` _but the chain id has no effect on transformed data_.

exactly - only the random seed is used to instantiate the model - https://github.com/stan-dev/cmdstan/blob/68517f0790e2a531e56f2a6ef3e323036abe5796/src/cmdstan/command.hpp#L225-L226
the chain id is passed in the the services method which then advance the RNG so that all chains are in different places in the sequence.



> (I don't understand what the footnote is talking about--it sure sounds like this it's very PR but it's been there for years...)

yes indeed - very PR and been there for years - I was involved in working on this maybe 2017 but all details escape me other than it was an issue for the old Stanc2 compiler and I worked on it.

betanalpha:
I’m going to second a design doc here.  I understand that the implementation is straightforward and will maintain the expected behavior when using a single seed and multiple chain ids, but as has been made clear from the long discussion here there’s enough confusion about how the Stan core services behave and how the interfaces have been implemented that the _context_ for this implementation needs to be carefully documented.

Part of the confusion here is a result from the growing technical debt arising from the consistent interface drift over the past few years.  One of the huge advantages of pulling multiple chains into the core services is that it ensures more uniform behavior across the interfaces (and less functionality that the interfaces have to maintain), but we need to make sure that everyone is on the same page of exactly what the current behavior is _supposed_ to be and what behavior this new functionality will enforce.  In other words the design doc is needed not for the implementation but for the API.

If anything a design doc will serve as a draft for better documentation moving forwards.

rok-cesnovar:
I kind of agree on a light design doc as well. The devs that have been following the process probably get the entire idea, but others would not. It does not have to be a full-fledged design doc. 

This hasnt been decided yet, but if we go with threading on by default, that is something that should be discussed more broadly not just something decided in a PR.

It should mention: 
- how it differs from the regular multi process cmdstan approach and what are the gains
- what to expect or not expect if `_rng` in transformed data
- what to expect with _rng in GQ and where does the actual seed come from (seed + chain ID)
- the potential thread local AD tape and lgamma performance regression that would come up if we go with threads on but only use 1 chain (This is potentially a non-issue, but we have not confirmed that yet).

> Because I also don't have access to a windows computer so idk how to do that benchmark.

If someone point me to what to run (model + data), I can run it easily.

SteveBronder:
Lol okay multiple people have also asked me on the side to make a design doc so I'm willing to say I'm wrong about the simplicity here and can write something up. A design doc should also make it easier for upstream folks so they have an implementation / API reference

SteveBronder:
Alrighty I put up a design doc below!

https://github.com/stan-dev/design-docs/pull/40

wds15:
Hmm.... when I do this:

```
examples/one_comp/one_comp sample num_samples=150 num_warmup=150 data file=examples/one_comp/one_comp.data.R parallel chains=8 threads=6
```

then only a single core is being used, but I would expect that 6 threads are being used at the same time.

When I now do

```
STAN_NUM_THREADS=6 examples/one_comp/one_comp sample num_samples=150 num_warmup=150 data file=examples/one_comp/one_comp.data.R parallel chains=8 threads=6
```

It crashes with

```
thread_monitor Invalid argument in pthread_attr_setstack_size
```

In any case, I do think that setting "threads" on the command line should always take precedence over `STAN_NUM_THREADS`.

SteveBronder:
Are you compiling cmdstan with `STAN_THREADS` on? If so I'll double check what's going on here

wds15:
Yes, I do.

SteveBronder:
@wds15 just updated the branch with the new Stan changes if you have time to give it a whirl!

wds15:
Looks like this:

```
examples/one_comp/one_comp sample num_samples=150 num_warmup=150 data file=examples/one_comp/one_comp.data.R parallel chains=8 threads=6
```

Does now what I want. I get 8 chains being run using 6 threads total. That's great. Need to check now what I am getting here...

wds15:
FYI, I just ran this:

```
./bernoulli sample data file=bernoulli.data.R threads=4 chains=4
```

and this did simply run a single chain, but it should raise an error, right?

This works:

```
./bernoulli sample chains=4 data file=bernoulli.data.R threads=4
```

... and how does this work out when the user sets `STAN_NUM_THREADS` in addition? We also need to check the display of the `num_threads` line in the cmdstan header.

SteveBronder:
Ah, yikes! This should be fixed you can now call 

```
./examples/bernoulli/bernoulli sample num_chains=4 data file=./examples/bernoulli/bernoulli.data.R num_threads=-1
```

> ... and how does this work out when the user sets STAN_NUM_THREADS in addition? We also need to check the display of the num_threads line in the cmdstan header.

That's a good Q, I tried setting some logic in here, can you check if it makes sense? Since the default of `STAN_NUM_THREADS` is 1 I ignore the env variable in that case.

wds15:
I just tried:

```
[21:28:24][sebi@sebastians-macbook-pro-1:~/work/cmdstan/examples/bernoulli]$ STAN_NUM_THREADS=2 ./bernoulli sample num_chains=4 data file=bernoulli.data.R num_threads=6
STAN_NUM_THREADS= 2 but argument num_threads= 6. Please either only set one or make sure they are equal.
```

There are extra spaces.... but my thinking here is that we actually should sample and use what is being passed to the argument `num_threads`. In addition we can emit a warning stating that things are not equal? In any case, I'd think we give priority to the `num_threads` argument would be my preference.

src/cmdstan/command.hpp
```diff
@@ -276,7 +276,19 @@ int command(int argc, const char *argv[]) {
 #ifdef STAN_THREADS
   n_threads = get_arg_val<int_argument>(parser, "num_threads");
 #endif
+  // Need to make sure these two ways to set thread # match.
```
wds15, line 279: What happens in line 277 whenever STAN_THREADS is not defined? In that case that line is not there, but the user may correctly specify num_threads=1 or is an error being raised if num_threads is given as argument?

src/cmdstan/arguments/arg_num_threads.hpp
```diff
@@ -10,8 +10,8 @@ class arg_num_threads : public int_argument {
   arg_num_threads() : int_argument() {
     _name = "num_threads";
     _description = std::string("Number of threads available to the program.");
-    _default = "-1";
-    _default_value = -1;
+    _default = "1";
+    _default_value = 1;
```
wds15, line 14: Can we encode here a minimal and maximal value of 1 whenever `STAN_THREADS` is absent (I am not familiar with the argument processing stuff too much; so this is a naive q)? Whenever `STAN_THREADS` is defined, then there is no upper limit. I recall that others were not happy with using -1 for "use all cores". Instead "0" was preferred, but we probably keep that consistent.

examples/bernoulli/bernoulli.stan
```diff
@@ -6,6 +6,10 @@ parameters {
   real<lower=0,upper=1> theta;
 }
 model {
-  theta ~ beta(1,1);  // uniform prior on interval 0,1
-  y ~ bernoulli(theta);
+  profile("gahh") {
```
wds15, line 9: clean me?!

examples/diamonds/diamonds.stan
```diff
@@ -0,0 +1,35 @@
+functions {
+}
+data {
+  int<lower=1> N;  // number of observations
```
wds15, line 4: Why do we add so many examples with this PR?

src/cmdstan/arguments/arg_num_threads.hpp
```diff
@@ -0,0 +1,22 @@
...
+ public:
+  arg_num_threads() : int_argument() {
+    _name = "num_threads";
+    _description = std::string("Number of threads available to the program.");
+    _default = "1";
+    _default_value = 1;
+    _value = _default_value;
+  }
+
+  bool is_valid(unsigned int value) { return value > -2; }
```
wds15, line 18: and it must never be equal to 0.

src/cmdstan/command.hpp
```diff
@@ -90,19 +96,143 @@ std::shared_ptr<stan::io::var_context> get_var_context(const std::string file) {
...
+using context_vector = std::vector<shared_context_ptr>;
+/**
+ * Make a vector of shared pointers to contexts.
+ * @param file The name of the file. For multi-chain we will attempt to find
+ *  {file_name}_1{file_ending} and if that fails try to use the named file as
+ *  the data for each chain.
+ * @param n_chain The number of chains to run.
+ * @return An std vector of shared pointers to var contexts
+ */
+context_vector get_vec_var_context(const std::string &file, size_t n_chain) {
```
wds15, line 114: `n_chain` -> `num_chain` ?!

src/cmdstan/command.hpp
```diff
@@ -90,19 +96,143 @@ std::shared_ptr<stan::io::var_context> get_var_context(const std::string file) {
...
+  using stan::io::var_context;
+  if (n_chain == 1) {
+    return context_vector(1, get_var_context(file));
+  } else {
+    auto make_context = [](auto &&file, auto &&stream,
```
wds15, line 119: else not needed, right?

src/cmdstan/command.hpp
```diff
@@ -90,19 +96,143 @@ std::shared_ptr<stan::io::var_context> get_var_context(const std::string file) {
...
+            throw std::invalid_argument(msg.str());
+          }
+          ret.push_back(make_context(file_i, stream_i, file_ending));
+        }
+        return ret;
+      }
+    }
+  }
+  // This should not happen
+  using stan::io::dump;
```
wds15, line 200: should not happen => throw an error?

wds15 (changes requested):
We need to settle on the priority thing for num_threads as argument vs STAN_NUM_THREADS in the environment. Other than that I think this is fine as far as I can see.

examples/bernoulli/bernoulli.stan
SteveBronder: Yeah sorry so fyi I just have these in here for easy testing locally but once the Stan PR is merged and we are about to merge this I'm going to delete all these examples

src/cmdstan/command.hpp
SteveBronder: So this is more of a weird C++ thing, we can't have the function reach the end without having a valid return. This code won't ever happen but this path does need to exist

src/cmdstan/command.hpp
SteveBronder: I just changed this so that the `num_threads` argument is always equal to 1 if `STAN_THREADS` is not defined. So I think now we should always be fine

SteveBronder:
> We need to settle on the priority thing for num_threads as argument vs STAN_NUM_THREADS in the environment. Other than that I think this is fine as far as I can see.

imo I think we should have `num_threads != STAN_NUM_THREADS` throw as I think the user should be aware of which is being set and how many threads they are giving to the program. I think the case of -1 and the `num_threads` being equal is fine as in that case the program will use the same number of threads.

wds15:
Now that stan is up to date we can finish this, right? Can u ping me, @SteveBronder , once ready, please?

SteveBronder:
Yes think this is ready for review now! Updated to the most recent stan version from develop and removed all the other stan programs in the examples folder. This is ready for review though I think it would also be good to have some tests in here that call the new multi chain interface. I'll try to get to those today but otherwise I can do them tmrw

wds15:
Bump. Tests are failing and I recall you wanted to add more tests... right? It would be cool to get this in now that the ground work is done. Let me know when it's my turn again.

SteveBronder:
@wds15 I gotta write one test and then do a quick lookover for cleanup and this should be good to go!

SteveBronder:
@wds15 we should be good to go! Two little things here

1. I had to modify the make files because when I turned threading on cmdstan was having a hard time compiling stansummary diagnostics etc.
2. I added a little test that just runs a simple test model with 2 chains. I think setting this up for doing threading would be a bit much for once test and we do that testing at the stan repo level so I think having num_chains go off is reasonable.

I think this is ready for review!

src/test/interface/multi_chain_test.cpp
```diff
@@ -0,0 +1,66 @@
...
+  model_path.push_back("test");
+  model_path.push_back("test-models");
+  model_path.push_back("test_model");
+
+  std::string command
+      = cmdstan::test::convert_model_path(model_path)
+        + " sample num_warmup=200 num_samples=1 num_chains=2 random seed=1234"
+        + " output file=" + cmdstan::test::convert_model_path(model_path)
+        + ".csv";
+
```
wds15, line 20: This test implicitly tests that the default init chain id is 1 (as we test that the filename ends with _1.csv)...do we want that? Or is it better to specify chain_id=1 as an argument to cmdstan? I don't know the answer, but some thought and doc would be nice here.

src/cmdstan/command.hpp
```diff
@@ -869,16 +1137,16 @@ int command(int argc, const char *argv[]) {
...
-          elbo_samples, max_iterations, tol_rel_obj, eta, adapt_engaged,
-          adapt_iterations, eval_elbo, output_samples, interrupt, logger,
-          init_writer, sample_writer, diagnostic_writer);
+          model, *(init_contexts[0]), random_seed, id, init_radius,
+          grad_samples, elbo_samples, max_iterations, tol_rel_obj, eta,
+          adapt_engaged, adapt_iterations, eval_elbo, output_samples, interrupt,
+          logger, init_writers[0], sample_writers[0], diagnostic_writers[0]);
     }
   }
   stan::math::profile_map &profile_data = get_stan_profile_data();
```
wds15, line 1152: The interaction of the multi chain stuff is probably not great with the profiling, right? At the moment you get a single profile.csv where multiple thread id's are listed.... or is this exactly what one wants? Maybe @rok-cesnovar has some thoughts? I'd guess we can go with this as it is presumably.

wds15 (changes requested):
If I run

`./bernoulli-2 sample num_chains=4 id=100 output refresh=1000 data file=bernoulli.data.R num_threads=4`

I'd expect to get `output_100.csv`, `output_101.csv` and so on...but I do get `output_1.csv`, `output_2.csv` and so on... I think we want the chain id to be the number, not 1...num_chains.

src/cmdstan/command.hpp
rok-cesnovar: I think a single profile CSV file is what I would want to see. Its less files to deal with and all the info is in one place.

As I said in previous discussions, profiling is something one would typically run with a single chain, as there is no added information gained when running multiple chains. It just adds more noise to the per-gradient numbers. 

So I would keep this simple.

src/test/interface/multi_chain_test.cpp
SteveBronder: Yes this was wrong before good catch! I set this to `id=10` so that in the test we then look for files ending in `10.csv` and `11.csv`. I think I should add docs to the `output` `file` argument so that folks know that for `id=N` and `num_chains=j` the output will be `output_{N + 1}.csv`, `output_{N + 2}.csv`, ..., `output_{N + j}.csv`

SteveBronder:
This is failing because of a recent change in stanc3. I opened up a revert for that so once that goes in we can rerun the tests and all should be good

https://github.com/stan-dev/stanc3/pull/945

SteveBronder:
@wds15 ready for another look!

wds15:
This error msg is a bit confusing (not sure if it can be improved easily):

```
[20:55:19][sebi@sebastians-macbook-pro:~/work/cmdstan/examples/bernoulli]$ ./bernoulli sample num_chains=4 id=10 data file=bernoulli.data.R output file=samples.csv num_threads=0
0 is not a valid value for "num_threads"
  Valid values: All
Failed to parse command arguments, cannot run model.
```

Valid values: All is obviously not true...

wds15:
And I think here is a bug:

```
./bernoulli sample num_chains=2 id=200 data file=bernoulli.data.R output file=samples.csv diagnostic_file=diag.csv num_threads=1
```

gives these files:

```
-rw-r--r--   1 sebi  staff    67483 23 Aug 21:03 diag_201les.csv
-rw-r--r--   1 sebi  staff    67531 23 Aug 21:03 diag_200les.csv
-rw-r--r--   1 sebi  staff    49336 23 Aug 21:03 samples_201.csv
-rw-r--r--   1 sebi  staff    49435 23 Aug 21:03 samples_200.csv
```

The "les" included in the diagnostic filename seems wrong?

SteveBronder:
I fixed up the error message so now we have 

```
0 is not a valid value for "num_threads"
  Valid values: num_threads > 0 || num_threads == -1
```

and fixed up the error message for `num_chains` as well to be more helpful. Also added a test to make sure the diagnostic file name is written correctly. Before I was doing a goof and appending the end of the sample file name to the diagnostic file name.

Issue #953, opened 2020-12-03
detect if model has no parameters and run fixed param sampler automatically


mitzimorris:
#### Summary:

When running the sampler on a model which doesn't have any parameters, the command program should be
able to detect whether or not the model has no parameters, and in this case, it should run fixed_param automatically.

#### Description:

It is extremely annoying to get the "fixed_param" error message when trying to use Stan programs to produce simulated data.

check here: https://github.com/stan-dev/cmdstan/blob/9b711ed1bd13db3d29c3df682333abe3813abb10/src/cmdstan/command.hpp#L390-L392

instead of quitting, the reasonable thing to do is run the non-adaptive sampler for the specified number of sampling iterations.

#### Reproducible Steps:

here's a datagen program in Stan:

```
// generate parameters, data, outcomes for poisson regression 
transformed data {
  int N_obs = 20;               // specify constants
}
generated quantities {
  int N = N_obs;
  int y_sim[N_obs];
  vector[N_obs] x_sim; 
  vector[N_obs] pop_sim;
  real alpha_sim;
  real beta_sim;
  real phi_sim;
  real mu[N_obs];  // mean

  alpha_sim = normal_rng(0, 1);  // simulate parameters
  beta_sim = normal_rng(0, 1);   // from prior
  while (1) {
    phi_sim = normal_rng(0, 1);
    if (phi_sim > 0) break;
  }

  for (n in 1:N_obs) {           // simulate data from params
    pop_sim[n] = uniform_rng(500,2500);
    x_sim[n] = uniform_rng(0.01,0.99);
    mu[n] = log(pop_sim[n]) + alpha_sim + x_sim[n] * beta_sim;
    y_sim[n] = neg_binomial_2_log_rng(mu[n], phi_sim);
  }
}
```

#### Current Output:
Run the Stan sampler - error message: "Must use algorithm=fixed_param for model that has no parameters."

#### Expected Output:
Describe what you expect the output to be. Knowing the correct behavior is also very useful.

Should just work.

#### Current Version:
v2.25.0

rok-cesnovar:
Closed this one by mistake.

PR #992, opened 2021-03-11
Print errors to stderr instead of stdout

mitzimorris:
#### Submisison Checklist

#### Summary:

- `command.hpp` and `main.cpp` send error messages to cerr instead of cout
-  when method `sample` is specified, if model doesn't contain any parameters, run algorithm `fixed_param` and print message to console.  https://github.com/stan-dev/cmdstan/issues/953
- slightly modified return code handling:  code `USAGE` used only for unrecognized command line args.  printing usage message returns code is `OK`.

#### Intended Effect:

make it easy for CmdStanX wrappers to report error messages via contents of stderr instead of making them scrape stdout.

#### How to Verify:

unit tests

#### Side Effects:

NA

#### Documentation:

NA

#### Copyright and Licensing


mitzimorris:
we'll need to advertise this change because folks might have scripts or packages that scrape stderr and/or stdout.

I don't quite understand why services layer uses return codes plus callback writers; effectively, there are only two return codes:  OK, not OK.   plus whatever messages are written to stderr.

src/cmdstan/command.hpp
```diff
@@ -125,12 +125,16 @@ int command(int argc, const char *argv[]) {
...
+    std::cerr << "Failed to parse command arguments, cannot run model."
+              << std::endl;
+    return err_code;
+  } else if (err_code != 0) {
+    std::cerr << "Unexpected failure, quitting." << std::endl;
     return err_code;
   }
   if (parser.help_printed())
-    return err_code;
+    return stan::services::error_codes::OK;
```
bbbales2, line 137: It looks like `rm` sets an error code here, so maybe we should too?

```
$ rm
rm: missing operand
Try 'rm --help' for more information.
$ echo $?
1
```

src/cmdstan/arguments/argument_parser.hpp
```diff
@@ -19,9 +19,12 @@ class argument_parser {
 
   int parse_args(int argc, const char *argv[], stan::callbacks::writer &info,
                  stan::callbacks::writer &err) {
+    if (argc == 0) {
+      return stan::services::error_codes::USAGE;
```
bbbales2, line 23: This should basically never happen right? Like `argc >= 1` cause the first argv will be the name of the program executing things?

bbbales2 (changes requested):
Question about the return code on printing error message. Change of behavior for zero parameter model makes sense.

src/cmdstan/arguments/argument_parser.hpp
mitzimorris: yup, should never happen - but there's  unit test for it - get rid of both?

src/cmdstan/command.hpp
mitzimorris: right, but this is more like the following:

```
~> make
make: *** No targets specified and no makefile found.  Stop.
~> echo $?
2
~> make -h
Usage: make [options] [target] ...
Options:
  -b, -m                      Ignored for compatibility.
 ....
~> echo $?
0
~> 
```

src/cmdstan/arguments/argument_parser.hpp
mitzimorris: removed those lines - not necessary.
leaving unit test alone.

src/cmdstan/command.hpp
mitzimorris: lines 136-137 are correct.  changed line 128 - return code `USAGE` set when no args are given - this will result in help message being printed; and then program will exit.  so if you get to line 136, user requested help to be printed, help was printed, all good.  cf above example

mitzimorris:
@bbbales2 - thanks for the code review - made changes, ready for re-review.

bbbales2 (approved):
Good!

bbbales2:
I updated the pull request name so it'd be easier to remember to add a warning about this to the release notes.

PR #994, opened 2021-03-14
All error messages written to stderr, including model instantiation failures; simplify return codes

mitzimorris:
#### Submisison Checklist

#### Summary:

Ensures that all error messages are written to stderr, including bad input data.
PR #992 overlooked the output stream passed in to the `new_model` function; changed it from standard output to standard error, as the only messages written to this output stream are error messages.

#### Intended Effect:

- Consistent handling of error messages - all error messages should go to standard error.

- Simplify return codes - the set of POSIX return codes in stan::services::error_codes aren't actionable for CmdStan users and the interfaces which wrap them.  

#### How to Verify:

Unit tests.

#### Side Effects:

As for #992, document that errors are not going to standard error, not standard out.

#### Documentation:

CmdStan User's guide now has section on return codes and common errors.

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company):  Columbia University

### DIFF
--- a/src/cmdstan/command.hpp
+++ b/src/cmdstan/command.hpp
@@ -226,7 +226,7 @@ int command(int argc, const char *argv[]) {
 
   // Instantiate model
   stan::model::model_base &model
-      = new_model(*var_context, random_seed, &std::cerr);
+      = new_model(*var_context, random_seed, &std::cout);
 
   std::vector<std::string> model_compile_info = model.model_compile_info();
 
--- a/src/cmdstan/command.hpp
+++ b/src/cmdstan/command.hpp
@@ -68,6 +68,13 @@ stan::math::profile_map &get_stan_profile_data();
 
 namespace cmdstan {
 
+struct return_codes {
+  enum {
+    OK = 0,
+    NOT_OK = 1
+  };
+};
+
 #ifdef STAN_MPI
 stan::math::mpi_cluster &get_mpi_cluster() {
   static stan::math::mpi_cluster cluster;
@@ -129,13 +136,13 @@ int command(int argc, const char *argv[]) {
     if (argc > 1)
       std::cerr << "Failed to parse command arguments, cannot run model."
                 << std::endl;
-    return err_code;
+    return return_codes::NOT_OK;
   } else if (err_code != 0) {
     std::cerr << "Unexpected failure, quitting." << std::endl;
-    return err_code;
+    return return_codes::NOT_OK;
   }
   if (parser.help_printed())
-    return stan::services::error_codes::OK;
+    return return_codes::OK;
 
   arg_seed *random_arg
       = dynamic_cast<arg_seed *>(parser.arg("random")->arg("seed"));
@@ -153,7 +160,7 @@ int command(int argc, const char *argv[]) {
       || (!opencl_device_id->is_default()
           && opencl_platform_id->is_default())) {
     std::cerr << "Please set both device and platform OpenCL IDs." << std::endl;
-    return stan::services::error_codes::CONFIG;
+    return return_codes::NOT_OK;
   } else if (!opencl_device_id->is_default()
              && !opencl_platform_id->is_default()) {
     stan::math::opencl_context.select_device(opencl_platform_id->value(),
@@ -219,7 +226,7 @@ int command(int argc, const char *argv[]) {
 
   // Instantiate model
   stan::model::model_base &model
-      = new_model(*var_context, random_seed, &std::cout);
+      = new_model(*var_context, random_seed, &std::cerr);
 
   std::vector<std::string> model_compile_info = model.model_compile_info();
 
@@ -252,7 +259,7 @@ int command(int argc, const char *argv[]) {
   std::shared_ptr<stan::io::var_context> init_context = get_var_context(init);
 
   // Invoke specified method
-  int return_code = stan::services::error_codes::OK;
+  int return_code = return_codes::OK;
 
   if (parser.arg("method")->arg("generate_quantities")) {
     string_argument *fitted_params_file = dynamic_cast<string_argument *>(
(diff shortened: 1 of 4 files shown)


mitzimorris:
@syclik - any thoughts w/r/t return codes?

src/cmdstan/command.hpp
```diff
@@ -219,7 +223,7 @@ int command(int argc, const char *argv[]) {
 
   // Instantiate model
   stan::model::model_base &model
-      = new_model(*var_context, random_seed, &std::cout);
+      = new_model(*var_context, random_seed, &std::cerr);
```
bbbales2, line 226: Will this make `print` statements in the Stan model go to stderr?

bbbales2 (changes requested):
Simplifying the error codes seems fine, though it sorta makes me wonder if we shouldn't go remove the return codes in services if they aren't being used.

One question about the output stream.

src/cmdstan/command.hpp
mitzimorris: excellent question.  will check.

mitzimorris: answer is no.  the generated model.hpp file contains the function `new_model`; this instantiates the model object.

the instantiated model object is passed into the services methods along with stderr and stdout, and therefore the correct output streams are hooked up to the calls to `log_prob` et al.

can't connect the c++ dots to understand why calls to `rethrow_located` in the model constructor go to stderr, but they do.

mitzimorris:
> we shouldn't go remove the return codes in services if they aren't being used.

in theory, sounds like a good idea - cf this nice writeup of why it's better to throw exceptions rather than use return codes: http://javierferrer.me/exceptions-vs-error-codes/

but we can't do this (yet) - my understanding is that RStan and PyStan call the services methods directly and they couldn't handle C++ calls throwing exceptions, so we went this way.  maybe @syclik can explain more.

src/cmdstan/command.hpp
mitzimorris: `print` statements in `transformed data` block will go to std err, because those are rolled into the model constructor.  everything else goes to stdout.

mitzimorris: also functions that print stuff will print to stderr when called in transformed data block, otherwise, will print to stdout.

rok-cesnovar: I do not think this should be moved to cerr as it will just be more confusing. 

As far as I understand, this stream for the constructor is not used anywhere apart from prints in transformed data. model instantiation failures are caught by https://github.com/stan-dev/stan/blob/develop/src/stan/io/validate_dims.hpp#L42

Log prob is passed the logger (https://github.com/stan-dev/cmdstan/blob/develop/src/cmdstan/command.hpp#L104).

If we need to change something in the codegen to make a proper solution for all of this or any larger refactor, lets do that. I  am against a quick fix that would also print users debug messages to stderr.

bbbales2: Sounds like this is a cleanup for stanc3. Since it's only transformed data prints, then I think revert this change and just do the return code cleanup here.

mitzimorris: the model instantiation failures caught by `validate_dims` are printed to whatever output stream is passed in to `new_model`.    I spent an hour yesterday trying to figure out why - I don't see an output stream passed around, and yet, that's where these error messages end up.

the result it that it's necessary to parse stdout to figure out if the data supplied to model is OK.  cf https://github.com/stan-dev/cmdstanpy/blob/2a7a9a5cc3a8e37670092c0af11d1e4c6d252381/cmdstanpy/stanfit.py#L230-L244. - this is a quick and dirty parse - it's brittle, might catch too much/too little, etc.

just to be clear, if the user has a `print` statement in transformed data block, it will go to stderr; otherwise, stdout.
but yes, agree, this is terrible.

bbbales2: Hmm I guess we can go ahead and fix this streaming thing then.

I was looking at the generated cpp for a model:

```cpp
stan::model::model_base& new_model(
        stan::io::var_context& data_context,
        unsigned int seed,
	std::ostream* msg_stream) {
  stan_model* m = new stan_model(data_context, seed, msg_stream);
  return *m;
}
```

Then I went to look at the stan_model constructor in my generated code and the pstream__ argument is unused there. Where does this argument come in to play anyway?

bbbales2: Or I guess the question is how does this argument end up changing where stuff goes anyway? Is this a problem with how some service function is not using loggers where it should be or something weird?

mitzimorris: I'd love to figure out what's going on here and then address it in a future PR.
for now,  can let thing go to stdout for now and just simplify return codes with this PR.

bbbales2 (approved):
Good!

mitzimorris:
changed title to match outcome of discussion.

syclik:
> but we can't do this (yet) - my understanding is that RStan and PyStan call the services methods directly and they couldn't handle C++ calls throwing exceptions, so we went this way. maybe @syclik can explain more.

I believe RStan and PyStan can handle the exceptions!

I agree that it makes more sense if they throw appropriate exceptions and let each interface handle error handling. In C++, if you don't handle an exception, it'll eventually crash the process. I'm not sure if it's well defined what the outcome is... it's better if we turned that exception into something reasonable (and returned a non-zero exit code).

syclik:
> when an uncaught exception causes the process to crash, depending on what the error is, the system returns a termination signal which is platform dependent, c.f this table: https://en.wikipedia.org/wiki/Signal_(IPC)#Default_action
> 
> understood that the calling process needs to handle the exception - this is always the case.

Thanks! Sorry about the language there... didn't mean to imply that you didn't understand that.

I was actually wondering what would happen if CmdStan didn't catch the exception and whether it'd be well-defined. Just from the perspective of what happens if we let it propagate (simplifies code)... I learned something new from your response... I didn't realize an uncaught exception results in a SIGTERM! I hadn't found that info in my searching around for what the behavior is.

mitzimorris:
definitely in favor of letting service methods throw errors instead of return codes.

PR #1009, opened 2021-05-07
Add datetime to CSV

### LINKED ISSUE #1006: Add timestamp to csv
#### Description:

This came up in https://github.com/stan-dev/cmdstanr/issues/490 but instead of doing this in only cmdstanr it would be nice if the CmdStan CSV files had a timestamp in them so it could be inferred when they were created. 

#### Current Version:

v2.26.1

rok-cesnovar:
#### Summary:

Adds the timestamp to the resulting CSV to close #1006

Example:
```
# stan_version_major = 2
# stan_version_minor = 26
# stan_version_patch = 1
# model = bernoulli_model
# start_time = 2021-05-07 20:09:04
# method = sample (Default)
```

#### Copyright and Licensing


src/cmdstan/write_datetime.hpp
```diff
@@ -0,0 +1,28 @@
...
+  std::tm* curr_tm = std::localtime(&current_datetime);
+  std::stringstream current_datetime_msg;
+  current_datetime_msg << "start_datetime = " << std::setfill('0')
+                       << (1900 + curr_tm->tm_year) << "-" << std::setw(2)
+                       << (curr_tm->tm_mon + 1) << "-" << std::setw(2)
+                       << curr_tm->tm_mday << " " << std::setw(2)
+                       << curr_tm->tm_hour << ":" << std::setw(2)
+                       << curr_tm->tm_min << ":" << std::setw(2)
+                       << curr_tm->tm_sec;
+  writer(current_datetime_msg.str());
```
rok-cesnovar, line 24: This is not pretty and can be replaced by 
```cpp
const std::time_t current_datetime = std::chrono::system_clock::to_time_t(std::chrono::system_clock::now());
std::stringstream current_datetime_msg;
current_datetime_msg << std::put_time(std::localtime(&current_datetime), "%F %T");
writer("start_datetime = " + current_datetime_msg.str());
```
once we up the minimum required g++ to 5+. g++ 4.9 does not have `std::put_time` even though its part of C++11...

I also tried using `std::strftime` and char arrays but that produced a few garbage characters.

rok-cesnovar:
Ready for review.

rok-cesnovar:
Last call for reviews.

wds15:
Can you give me today?

rok-cesnovar:
No one is racing to review this so take your time :)

wds15:
I see no tests. Do we say that comments like these are not required to be tested? What would happen downstream if the date-time is not present or messed up?

wds15:
And would it make sense to add the end time as well?

wds15:
And what about time-zone?

rok-cesnovar:
> I see no tests

Yeah, I am not sure what can we test in Cmdstan without doing some sort of parsing in Python. There will be tests in cmdstanr if that helps. We also dont tests the `num_threads` comments etc.

> What would happen downstream if the date-time is not present or messed up?

This does not touch or change anything in downstream (Math/Stan services) sonothing. And CSV readers in upstream (rstan, cmdstanr/py) do not care about the number of comments or position nor should they.

> And would it make sense to add the end time as well?

Hm, maybe, though that can also be obtained from start time + running time. The problem is that the way CSV is printed that would add another comment after the CSV and we are trying to avoid that as the format

```
# comments
values
# more comments
```
or even
```
# comments
values
# adaptiation results
more value
# comments
```
is causing problems for almost all fast CSV readers in R and Python and they need to be handheld a bit.

> And what about time-zone?

Hm, good call. Will add a "UTC+3", "UTC-2".

wds15:
What about stansummary printing out the date time... and that we test?

rok-cesnovar:
That means touching stansummary and I really dont want to do exapand stansummary for no reason except to test. Will try to add a test to parse in C++.

rok-cesnovar:
I fixed it to the UTC timezone and added a test. It checks that the start_datetime was found, that the year matches and that the start_datetime ends with "UTC".

rok-cesnovar:
If this PR is merged I would just proceed with the release.

wds15:
Looking at it now.

wds15 (approved):
LGTM. Thanks for adding the test.

rok-cesnovar:
Thanks for the quick review.

PR #1019, opened 2021-05-30
Add make/local.example and minor fixes

### LINKED ISSUE #942: make/local template
#### Summary:

I think it would be good to template `make/local` with options. Like how linux config files are these days. So if someone goes in there looking for stuff they see:

```
# Uncomment to enable threading
# STAN_THREADS=TRUE
```

And similarly for MPI, OpenCL, the common C++ flags, etc.

#### Current Version:
v2.25.0

rok-cesnovar:
#### Summary:

A bit of last minute cleaning pre 2.27:
- adds docs for STAN_NO_RANGE_CHECKS to `make help`
- uses new array syntax in the example model
- add the make/local.example file to showcase examples on how to use the make/local file (fixes #942)

#### Copyright and Licensing

Rok Češnovar

### DIFF
--- a/make/local.example
+++ b/make/local.example
@@ -1,4 +1,5 @@
-# This file contains examples for makefile flags that can be used in Cmdstan's make/local file
+# To use this template, make a copy from make/local.example to make/local and uncomment options as needed.
+# Be sure to run `make clean-all` before compiling a model to make sure everything gets rebuilt.
 
 # Change the C++ compiler
 # CXX=clang++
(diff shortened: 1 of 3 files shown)


make/local.example
```diff
@@ -0,0 +1,29 @@
...
+# STANCFLAGS+= --warn-pedantic
+
+# Enable C++ compiler and linker optimization recommended by Stan developers
+# STAN_CPP_OPTIMS=true
+
+# Remove range checks from the model for faster runtime
+# STAN_NO_RANGE_CHECKS=true
+
+# Adding other arbitrary C++ compiler flags
+# CXXFLAGS+= -march=native -mtune=native
```
bbbales2, line 28: [optional] The march/mtune errors that hit rstan was due to old toolchains? Any chance using these exact things as examples leads to headache?

make/local.example
```diff
@@ -0,0 +1,29 @@
+# This file contains examples for makefile flags that can be used in Cmdstan's make/local file
```
bbbales2, line 1: [optional] Making the directions on using this more explicit might be helpful.

```
# To use this template, make a copy from make/local.example to make/local and uncomment options as needed. Be sure to run `make clean-all` before compiling a model to make sure everything gets rebuilt.
```

Something like that

bbbales2 (approved):
This looks good. Couple optional things.

make/local.example
rok-cesnovar: Those issues are probably mostly gone, but just to be safe I will add something that is less "dangerous".

wds15: …and maybe make a note that the range checks should only be disabled for the final production run models?

rok-cesnovar: That was added in the docs here: https://github.com/stan-dev/docs/pull/365/files#diff-2eaed6ee2580309bec34cc5ba3cbe013877710c1efb49906978c9ea9eee33147R164 

But I guess it doesnt hurt to be safe yes. Its not like anyone reads the docs first and then jumps to use stuff. Myself included :)

bbbales2 (approved):
Good

PR #1032, opened 2021-08-12
Complex numbers vals_c

nickdidio:
#### Submisison Checklist

#### Summary:

Allow for intake of complex numbers in Stan. This PR mainly adds a new function called vals_c which allow for intake of complex numbers in a Stan program.

#### Documentation

All documentation for complex number is available at stan-dev/docs#378.

#### How to Verify:

Added more tests in /src/test/interface/io/json/json_data_test.cpp

Run tests: `./runCmdStanTests.py src/test`

#### Side Effects:

None

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company): Nicholas DiDio


src/cmdstan/io/json/json_data.hpp
```diff
@@ -119,6 +120,43 @@ class json_data : public stan::io::var_context {
     return empty_vec_r_;
   }
 
+  /**
+   * Read out a vector of complex numbers from the specifed name.
+   * @param name Name of Variable.
```
bob-carpenter, line 125: This needs `@return` doc.

It should also have some light doc on the presumed input format or a pointer to where it's described.  Otherwise, the code's too cryptic.

src/cmdstan/io/json/json_data.hpp
```diff
@@ -119,6 +120,43 @@ class json_data : public stan::io::var_context {
...
+   */
+  std::vector<std::complex<double>> vals_c(const std::string &name) const {
+    if (contains_r_only(name)) {
+      auto &&vec_r = (vars_r_.find(name)->second);
+      auto &&val_r = vec_r.first;
+      auto &&dim_r = vec_r.second;
+      std::vector<std::complex<double>> vec_c(val_r.size() / 2);
+      int offset = 1;
+      for (int i = 0; i < dim_r.size() - 1; ++i) {
```
bob-carpenter, line 134: [question]
why is this less than size - 1?  Don't you want to take the product of all of the elements? '

[optional]
You can convert the initialization and loop into a one-liner functional form using `std::accumulate`.

src/cmdstan/io/json/json_data.hpp
```diff
@@ -119,6 +120,43 @@ class json_data : public stan::io::var_context {
...
+        offset *= dim_r[i];
+      }
+      for (int i = 0; i < vec_c.size(); ++i) {
+        vec_c[i] = std::complex<double>{static_cast<double>(val_r[i]),
```
bob-carpenter, line 138: Why does `val_r[i]` need to be cast here?   Isn't it already a `double` value?  And even if it's not a double value, calling the `std::complex<double>` constructor should automatically cast it to `double` because of the template specification.

src/test/interface/io/json/json_data_test.cpp
```diff
@@ -52,6 +54,27 @@ void test_real_var(cmdstan::json::json_data &jdata, const std::string &text,
     EXPECT_EQ(expected_vals[i], vals[i]);
 }
 
+void test_complex_var(cmdstan::json::json_data &jdata, const std::string &text,
+                      const std::string &name,
+                      const std::vector<std::complex<double>> &expected_vals,
+                      const std::vector<size_t> &expected_dims) {
+  EXPECT_EQ(true, (jdata.contains_r(name) || jdata.contains_i(name)));
+  std::vector<size_t> dims = jdata.dims_r(name);
+  if (dims.size() == 1) {
```
bob-carpenter, line 63: You don't need this first conditional.  `pop_back()` will be equivalent to `clear()` if there is only a single element in the container.  You generally don't want to branch for this kind of thing as branching is very inefficient when there's branch-point misprediction.

src/test/interface/io/json/json_data_test.cpp
```diff
@@ -52,6 +54,27 @@ void test_real_var(cmdstan::json::json_data &jdata, const std::string &text,
...
+    dims.clear();
+  } else {
+    dims.pop_back();
+  }
+  EXPECT_EQ(expected_dims.size(), dims.size());
+  for (size_t i = 0; i < dims.size(); i++)
+    EXPECT_EQ(expected_dims[i], dims[i]);
+  std::vector<std::complex<double>> vals = jdata.vals_c(name);
+  ;
```
bob-carpenter, line 72: Remove the stray semicolon.  It's a valid empty statement, but it's not doing anything.

src/test/interface/io/json/json_data_test.cpp
```diff
@@ -144,6 +177,18 @@ TEST(ioJson, jsonData_real_array_1D) {
   test_real_var(jdata, txt, "foo", expected_vals, expected_dims);
 }
 
+TEST(ioJson, jsonData_complex_array_1D) {
```
bob-carpenter, line 180: [question]
Is there a test that we get the right error message if the variable is not found?

bob-carpenter (changes requested):
This looks great.  The requested changes are all minor.

src/cmdstan/io/json/json_data.hpp
nickdidio: Not sure why we're doing size - 1. Steve wrote this code and might have the answer for this.

src/test/interface/io/json/json_data_test.cpp
nickdidio: We test if the variable name exists in test_complex_var

`EXPECT_EQ(true, (jdata.contains_r(name) || jdata.contains_i(name)));`

src/cmdstan/io/json/json_data.hpp
bob-carpenter: Coming back to it, I now see that it's just multiplying out first N - 1 dimension sizes so it can index into the last dimension.  So this is fine.

bob-carpenter (approved):
Looks good.  Thanks for updating.

PR #1038, opened 2021-09-11
Fix disabling precompiled headers properly

rok-cesnovar:
#### Summary:

With clang, PRECOMPILED_HEADERS=false currently has no effect, the .hpp.gch files are still built. This fixes that problem.

#### Copyright and Licensing

### DIFF
--- a/make/program
+++ b/make/program
@@ -28,7 +28,7 @@ $(STAN)src/stan/model/model_header$(STAN_FLAGS).hpp.gch : $(STAN)src/stan/model/
 
 ifeq ($(CXX_TYPE),clang)
 CXXFLAGS_PROGRAM += -include-pch $(STAN)src/stan/model/model_header$(STAN_FLAGS).hpp.gch
-$(STAN_TARGETS) examples/bernoulli/bernoulli$(EXE) $(patsubst %.stan,%$(EXE),$(wildcard src/test/test-models/*.stan)) : %$(EXE)
+$(STAN_TARGETS) examples/bernoulli/bernoulli$(EXE) $(patsubst %.stan,%$(EXE),$(wildcard src/test/test-models/*.stan)) : %$(EXE) : $(STAN)src/stan/model/model_header$(STAN_FLAGS).hpp.gch
 endif
 endif
 
--- a/make/program
+++ b/make/program
@@ -18,6 +18,8 @@ $(CMDSTAN_MAIN_O) : $(CMDSTAN_MAIN)
 ##
 $(STAN)src/stan/model/model_header$(STAN_FLAGS).d : $(STAN)src/stan/model/model_header.hpp
 	$(COMPILE.cpp) $(DEPFLAGS) $<
+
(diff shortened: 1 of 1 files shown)


rok-cesnovar:
You can validate this by running

```
make build -j4
make examples/bernoulli/bernoulli
```
first with an empty make/local. The file `stan/src/stan/model/model_header.hpp.gch` should be there.

Then set `PRECOMPILED_HEADERS=false` and run

```
make clean-all
make build -j4
make examples/bernoulli/bernoulli
```

WardBrian:
What makes this clang-specific? I believe the flag worked correctly for me when setting up the MacOS conda installation

rok-cesnovar:
Yeah, its not clang-specific. I fixed the title. 

> I believe the flag worked correctly for me when setting up the MacOS conda installation

Do you mean disabling precompiled headers worked?

WardBrian:
Yes. [This commit](https://github.com/conda-forge/staged-recipes/pull/15203/commits/29b00ecde151b4a2137d8003231fa32b57df3d69) was the first which passed their CI on MacOS

rok-cesnovar:
Oh yeah, that will make it not build on "make build":

https://github.com/stan-dev/cmdstan/blob/develop/makefile#L117 and https://github.com/stan-dev/cmdstan/blob/develop/makefile#L247

It did still include it here when you compiled a model: https://github.com/stan-dev/cmdstan/blob/develop/make/program#L29
And that made make build .hpp.gch files then.

WardBrian:
Oh, I think in the conda use case that's actually the best case scenario. The issue was if the PCH was compiled before install (e.g, when it was compiled on a different machine), but compiling it on first model compilation is good for later performance. 

So this would disable that?

rok-cesnovar:
Huh, yeah :/

I am not familiar with conda, but you could build cmdstan and delete the model_header.hpp.gch file with PRECOMPILED_HEADERS on. But I am not sure if that works for you.

rok-cesnovar:
What we could do as an alternative is not build the precompiled header on `make build` but always with the first model compilation. That would mean you need to remove the PRECOMPILED_HEADERS=false line.

WardBrian:
I suppose it's a question of whether we want the intended behavior to be as it currently is (the flag is only for `make build`) or if it should never build the pch file. I can probably make the conda build work as you suggest, I wonder if others are assuming this behavior however.

rok-cesnovar:
There is no real difference for users I guess. It only makes the installation more transferable. I dont see a use case of why it would matter that the PCH file is built on `make build`.

WardBrian:
The CI is finishing up now but it seems like deleting the file as part of the conda build is just as good, so either change is fine

rok-cesnovar:
Ok, cool. Lets leave it as is then if it works for you.

WardBrian (approved):

rok-cesnovar:
Thanks Brian!

PR #1039, opened 2021-09-16
Precompile windows gcc 8

rok-cesnovar:
#### Summary:

This PR changes the makefile logic so that `PRECOMPILED_HEADERS=true` is automatically set on Windows if its using g++ 8 or newer (RTools 4.0 supplied toolchain).
Currently its not used on Windows by default.

#### Copyright and Licensing

Rok Češnovar

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -32,6 +32,7 @@ jobs:
       shell: powershell
     - name: Build Math libs & add to PATH
       run: |
+        make print-PRECOMPILED_HEADERS
         mingw32-make -f stan/lib/stan_math/make/standalone math-libs
         echo "D:/a/cmdstan/cmdstan/stan/lib/stan_math/lib/tbb" | Out-File -Append -FilePath $env:GITHUB_PATH -Encoding utf8
       shell: powershell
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -32,6 +32,7 @@ jobs:
       shell: powershell
     - name: Build Math libs & add to PATH
       run: |
+        make print-PRECOMPILED_HEADERS
```
rok-cesnovar, line 35: This is just a test that shows that its still disabled in the Github Action test with Rtools 3.5.

makefile
```diff
@@ -108,7 +108,11 @@ endif
 STAN_FLAGS=$(STAN_FLAG_THREADS)$(STAN_FLAG_MPI)$(STAN_FLAG_OPENCL)
 
 ifeq ($(OS),Windows_NT)
+ifeq ($(shell expr $(CXX_MAJOR) \>= 8), 1)
```
SteveBronder, line 111: Should we also be checking they are using gcc and not clang or would it be fine in that case as well?

SteveBronder (approved):
One quick Q but looks good!

makefile
rok-cesnovar: Clang on windows with cmdstandoes not work at all (https://github.com/stan-dev/cmdstan/issues/736) so that didnt cross my mind. Will add a note there if we ever figure that out. 

Other option would be to keep the precompiled headers off in the off chance someone tries this with clang on Windows?

rok-cesnovar: Changed the logic to also handle clang on Windows. As we have no idea if it works, I opted to disabling it by default.

PR #1107, opened 2022-07-18
Add method for calculating log_prob and its gradients

andrjohns:
#### Submisison Checklist

#### Summary:

A common request is to add functionality comparable to `rstan`'s [`log_prob` and `grad_log_prob`](https://mc-stan.org/rstan/reference/stanfit-method-logprob.html) functions.

This PR adds an additional method to the command-line: `log_prob`, which takes an input file of (unconstrained) parameter values (`json` or `csv`) with variable(s) `params_r` and optionally `params_i` (as defined by the model's `log_prob` function), and returns a `.csv` file with the log-probability and gradients with respect to each parameter.

As an example, consider the bernoulli example with two parameters:
```stan
data {
  int<lower=0> N;
  array[N] int<lower=0,upper=1> y; // or int<lower=0,upper=1> y[N];
}
parameters {
  real<lower=0,upper=1> theta;
  real<lower=0,upper=1> theta2;
}
model {
  theta ~ beta(1,1);  // uniform prior on interval 0,1
  theta2 ~ beta(1,1);  // uniform prior on interval 0,1
  y ~ bernoulli(theta);
}
```

To use the `log_prob` method a file with parameter values is created (called `params.json` hereafter), with contents:
```
{"params_r":[1,0.5]} 
```

After the model has been compiled, the `log_prob` method is called as:
```
./bernoulli method=log_prob unconstrained_params=params.json data file=bernoulli.data.json output grad_output_file=grad.csv
```

The `grad.csv` file then contains:
```
# log_prob
-14.2073
# log_prob_grad
-5.7727,-0.244919
```

Note that this is my first foray into the IO side of things, so make sure to let me know if there's a more efficient way of handling the reading/writing here!

#### Intended Effect:

Simple access to the model's log-probability and its gradients for end users

#### Side Effects:

N/A

#### Copyright and Licensing


rok-cesnovar:
Love it! Works great!

It would be great if we were able to evaluate a bunch of log_probs with a single model instantiation. Not necessarily required to do here, but we should set up the API so that it allows us to expand to that in the future.

I think a better name for `grad_output_file` would be `log_prob_output_file`? Though I think we should separate the output into two files - splitting CSV files based on comments is a tedious job. Especially if we are able to have more than 1 row if we expand to multiple evals at some point.

nhuurre:
Is that `grad.csv` a well-formed CSV file? I'd rather have something like
```
lp__, g_theta, g_theta2
-14.2073, -5.7727, -0.244919
```
(`g_*` is how the gradients are named in `diagnostic_file`.)

The input being _unconstrained_ parameters is a bit problematic when the constraining/unconstraining transform is not exposed AFAIK. Could the API looked like `generate_quantities`, i.e. takes in a CSV of _constrained_ parameters?

bob-carpenter:
I was about to make @nhuure's proposal, then I scrolled down.  That will be *much* easier to ingest into a data analysis package as it's just a CSV file with a meaningful header.

Another request would be to do this for more than one input.  That is, take something that looks like a CSV file of draws and use that.  e reason I ask this is that it will be very slow to run this one gradient at a time because of the need to reload data and recompute transformed data.  This would wind up looking a lot like standalone generated quantities.  In fact, you could even add this to the standalone generated quantities functionality as additional columns of output.
the output and run it to calculate gradients.

bob-carpenter:
Yes, a server could be added to CmdStan.  I just didn't want to get involved with Stan governance in writing my own server because I've been overruled on all the other interface design decisions and want to maintain control of this one.

WardBrian:
I think that makes sense, since they would necessarily have different goals. I think anything incorporated into cmdstan would want to stay away from the binary IO which stan-model-server will eventually need for efficiency, for instance.

andrjohns:
As some final changes I've added some code comments, as well as an additional argument for whether or not to apply jacobian-adjustments when calculating log_prob (defaulting to TRUE).

That's the last of the changes that I can think of, so this is now ready for review

src/cmdstan/arguments/arg_log_prob.hpp
```diff
@@ -0,0 +1,26 @@
...
+#include <cmdstan/arguments/categorical_argument.hpp>
+
+namespace cmdstan {
+
+class arg_log_prob : public categorical_argument {
+ public:
+  arg_log_prob() {
+    _name = "log_prob";
+    _description
+        = "Return the log-probability and its gradients, given supplied "
```
bob-carpenter, line 16: log-probability -> log density up to a constant

P.S.  The term `log_prob` was an unfortunate choice based on BDA's use of "probability function" to refer to either a probability density function or probability mass function.  It's technically a mixed probability function because some of the variables can be discrete, but we use "density" as the general term for that (in retrospect, I think "_lpdf" and "_lpmf" distinctions were a mistake because with tuples we'll be able to have probability functions that are mixed.

src/cmdstan/arguments/arg_log_prob.hpp
```diff
@@ -0,0 +1,26 @@
...
+          "parameters";
+
+    _subarguments.push_back(new arg_log_prob_unconstrained_params());
```
bob-carpenter, line 19: to use move semantics, replace `push_back` with `emplace_back` and wrap args with `std::forward`

src/cmdstan/arguments/arg_log_prob_constrained_params.hpp
```diff
@@ -0,0 +1,24 @@
...
+
+#include <cmdstan/arguments/singleton_argument.hpp>
+
+namespace cmdstan {
+
+class arg_log_prob_constrained_params : public string_argument {
+ public:
+  arg_log_prob_constrained_params() : string_argument() {
+    _name = "constrained_params";
+    _description = "Input file of parameter values on constrained scale";
```
bob-carpenter, line 12: This needs doc on how the file is encoded.

src/cmdstan/arguments/arg_log_prob.hpp
```diff
@@ -0,0 +1,26 @@
...
+#define CMDSTAN_ARGUMENTS_ARG_LOG_PROB_HPP
+
+#include <cmdstan/arguments/arg_log_prob_unconstrained_params.hpp>
+#include <cmdstan/arguments/arg_log_prob_constrained_params.hpp>
+#include <cmdstan/arguments/arg_log_prob_jacobian_adjust.hpp>
+#include <cmdstan/arguments/categorical_argument.hpp>
+
+namespace cmdstan {
+
+class arg_log_prob : public categorical_argument {
```
bob-carpenter, line 11: Classes and constructors should get doc.

src/cmdstan/arguments/arg_log_prob_unconstrained_params.hpp
```diff
@@ -0,0 +1,24 @@
...
+
+#include <cmdstan/arguments/singleton_argument.hpp>
+
+namespace cmdstan {
+
+class arg_log_prob_unconstrained_params : public string_argument {
+ public:
+  arg_log_prob_unconstrained_params() : string_argument() {
+    _name = "unconstrained_params";
+    _description = "Input file of parameter values on unconstrained scale";
```
bob-carpenter, line 12: please include a short format description (like "JSON" or "R dump") in `_description`

src/cmdstan/command.hpp
```diff
@@ -594,6 +594,149 @@ int command(int argc, const char *argv[]) {
...
+    string_argument *upars_file = dynamic_cast<string_argument *>(
+        parser.arg("method")->arg("log_prob")->arg("unconstrained_params"));
+    string_argument *cpars_file = dynamic_cast<string_argument *>(
+        parser.arg("method")->arg("log_prob")->arg("constrained_params"));
+    bool jacobian_adjust
+        = dynamic_cast<bool_argument *>(
+              parser.arg("method")->arg("log_prob")->arg("jacobian_adjust"))
+              ->value();
+    if (upars_file->is_default() && cpars_file->is_default()) {
+      msg << "No input parameters provided, cannot calculate log-probability";
```
bob-carpenter, line 607: log-probability -> log probability

P.S.  The ridiculous rule of English punctuation is that you hyphenate noun compounds used as adjectives (e.g., "log-probability calculation") but not noun compounds used as nouns (e.g. "we calculate the log probability").

Also, this should say "log probability density".

src/cmdstan/command.hpp
```diff
@@ -594,6 +594,149 @@ int command(int argc, const char *argv[]) {
...
+      By only updating the sizes/vectors if there is data present, the
+      calculation and printing can be agnostic from the combinations of
+      un-/constrained parameters passed
+    */
+    size_t u_params_vec_size = 0;
+    size_t u_params_size = 0;
+    size_t c_params_vec_size = 0;
+    size_t c_params_size = 0;
+
+    std::vector<double> u_params_r;
```
bob-carpenter, line 621: declare these as low as possible so it's easier to line up where they're defined and where they're set;  for example, `c_params` should move lower to above where `c_params are sorted out.  I'd remove space in between declaration and conditional defining so the code proximity makes it clear what's defining them

src/cmdstan/command.hpp
```diff
@@ -594,6 +594,149 @@ int command(int argc, const char *argv[]) {
...
+
+      c_params_r = (*cpars_context).vals_r("params_r");
+      dims_c_params_r = (*cpars_context).dims_r("params_r");
+
+      c_params_vec_size = dims_c_params_r.size() == 2 ? dims_c_params_r[0] : 1;
+      c_params_size = dims_c_params_r.size() == 2 ? dims_c_params_r[1]
+                                                  : dims_c_params_r[0];
+    }
+
+    /*
```
bob-carpenter, line 659: all these need to be `//`.  Thanks!

src/cmdstan/command.hpp
```diff
@@ -594,6 +594,149 @@ int command(int argc, const char *argv[]) {
...
+    Parameter names and dimensions required for constructing
+    array_var_context and subsequently transforming constrained parameters
+    to unconstrained scale
+    */
```
bob-carpenter, line 663: shorten these doc strings---the names make it pretty clear already, and long doc strings get in the way of reading code.  For example, this one can be a one-liner:

```
store names and dims for constructing array_var_context and unconstraining transform
```

and the second one can be

```
store in single nested array to allow single loop for calc and print
```

The comments like this don't need to be long grammatical sentences---they're just for devs and the shorter the better.

src/cmdstan/command.hpp
```diff
@@ -594,6 +594,149 @@ int command(int argc, const char *argv[]) {
...
+      are passed, as the var_context returns them appended by parameter, rather
+      than by set, i.e,:
+        [[a1, b1],[a2,b2]] => [a1, a2, b1, b2]
+    */
+    using StrideT = Eigen::Stride<1, Eigen::Dynamic>;
+    for (size_t i = 0; i < c_params_vec_size; i++) {
+      Eigen::Map<Eigen::VectorXd, 0, StrideT> map_r(
+          c_params_r.data() + i, c_params_size, StrideT(1, c_params_vec_size));
+
+      // Constrained parameters are unconstrained directly into the std::vector
```
bob-carpenter, line 688: Remove this doc---this is just saying what `transform_inits` does.  If transform_init's doc isn't sufficient, add it there.

src/cmdstan/command.hpp
```diff
@@ -594,6 +594,149 @@ int command(int argc, const char *argv[]) {
...
+      stan::io::array_var_context context(param_names, map_r, param_dimss);
+      model.transform_inits(context, dummy_params_i, params_r_ind[i], &msg);
+    }
+
+    for (size_t i = c_params_vec_size; i < num_par_sets; i++) {
+      size_t iter = i - c_params_vec_size;
+      Eigen::Map<Eigen::VectorXd, 0, StrideT> map_r(
+          u_params_r.data() + iter, u_params_size,
+          StrideT(1, u_params_vec_size));
+      // As they are already on the correct scale, the unconstrained parameters
```
bob-carpenter, line 699: also, this comment is really just repeating the code, so I'd remove it

The goal is readable code, not heavily commented code!  As the Google C++ style guide says, only use code comments as a last resort when there's no way to write the code itself cleanly.  Here, I think the code itself is clean enough.  We don't want to document functions that are getting called---you should assume the person reading this code knows the Stan API.  Otherwise, it's never clear how much to doc of a Stan call.  

And sorry for being so pedantic if this is obvious.

src/cmdstan/command.hpp
```diff
@@ -594,6 +594,149 @@ int command(int argc, const char *argv[]) {
...
+                                                     dummy_params_i, gradients);
+      } else {
+        lp = stan::model::log_prob_grad<false, false>(
+            model, params_r_ind[i], dummy_params_i, gradients);
+      }
+
+      output_stream << lp << ",";
+
+      // Ensure that each row of values ends with a newline instead of delimeter
+      if (gradients.size() > 1) {
```
bob-carpenter, line 730: Just use the general case for both.  I assume the `ostream_iterator` will do the right thing with sizes 0 and 1 and it's not worth making the code more complicated for something like this.  And then the `<< '\n'` can be on the same line (it can be a character instead of string).

src/cmdstan/command.hpp
```diff
@@ -594,6 +594,149 @@ int command(int argc, const char *argv[]) {
...
+        std::copy(gradients.begin(), gradients.end() - 1,
+                  std::ostream_iterator<double>(output_stream, ","));
+        output_stream << gradients.back();
+      } else {
+        output_stream << gradients[0];
+      }
+      output_stream << "\n";
+    }
+    output_stream.close();
```
bob-carpenter, line 739: What happens if there's an exception somewhere in the I/O?  I'm used to seeing a `try/catch` to guarantee we close all the streams we open.  On the other hand, cmdstan just exits when there's a failure, so I doubt there's going to be any resource leakage past a failed call.  But it would be nice to include error messages.  is that handled somewhere else?

src/cmdstan/command.hpp
```diff
@@ -594,6 +594,149 @@ int command(int argc, const char *argv[]) {
...
   } else if (user_method->arg("diagnose")) {
     list_argument *test = dynamic_cast<list_argument *>(
         parser.arg("method")->arg("diagnose")->arg("test"));
```
bob-carpenter, line 742: Tests?

bob-carpenter (changes requested):
Thanks.  This looks great---these are almost all stylistic comments, so I hope they're not too annoying.

What's the situation for testing these commands?  I didn't see any tests anywhere that arguments are being handled correctly, etc.

src/cmdstan/arguments/arg_log_prob.hpp
andrjohns: I've added doc for the others, but I think it might be redundant to add doc here. The description covers what the method does, and the sub-arguments have doc for their purpose and type

andrjohns:
Thanks @bob-carpenter! I've updated the doc and comments, and added tests. Let me know if there are any aspects/functionalities that I've forgotten to test as well, thanks!

bob-carpenter (approved):
Thanks.  the new doc looks great.  I would've preferred to go ahead and doc everything, because then we don't need to make judgement calls on case-by-case exceptions.  But I don't want to hold this up further with doc, so I'll approve and merge.

WardBrian:
After updating my local checkout I've been trying to use this but most things I've tried so far have segfaulted. I'm not sure if I'm doing something wrong, the arguments need more validation checks, or something else. 

Using the standard bernoulli example, I've tried 

```
./bernoulli log_prob constrained_params=sample.csv data file=bernoulli.data.json 
```
where sample.csv was produced by sampling; segfault. I also tried this with the same CSV but removing all the columns except for `theta` and all the comment rows; still segfaults.

```
./bernoulli log_prob constrained_params=test.json data file=bernoulli.data.json 
```
Where test.json was `{"theta": 0.2}`; segfault
Where test.json was `{"theta": [0.4,0.5,0.2]}`; segfault. 


I'd like to know 1) what I'm doing wrong, and 2) can we try to catch whatever that is earlier

andrjohns:
Ah that's because the `constrained_params` needs to have a [single variable `params_r`](https://github.com/stan-dev/cmdstan/blob/1e2177540689147b5ec069d629e611f4c093b90c/src/cmdstan/arguments/arg_log_prob_constrained_params.hpp#L7), which is what the model's `log_prob` function takes

andrjohns:
There's an example file in the tests: https://github.com/stan-dev/cmdstan/blob/develop/src/test/test-models/bern_constrained_params.R

WardBrian:
Why wouldn't constrained_params accept something in the same format as the `inits` argument for sampling? E.g. using the actual parameter names and shapes?

Either way, this shouldn't segfault since I expect it will be a common thing people will try to do

andrjohns:
That's fair enough. I was matching the [rstan implementation](https://mc-stan.org/rstan/reference/stanfit-method-logprob.html), since I was aiming to add similar functionality for `cmdstanr`, but I can change it to take inputs matching the sample output, I'll do another PR

WardBrian:
I'm thinking from the standpoint of CmdStanPy etc, we will probably want to accept the constrained parameters as a dictionary. 

Instead taking in a `params_r` vector would be a lot of work for users and probably a source of subtle bugs - e.g. re-ordering declarations in the parameters block would make old inputs now return different outputs, which isn't the case if we do it all with the real names

WardBrian:
Looking at the rstan documentation, I believe they handle constrained parameters as a named list and unconstrained as a vector, rather than both as a vector.

mitzimorris:
updated PR description - input files are json or rdump, not CSV.

PR #1108, opened 2022-07-21
Update log_prob method to take constrained_params in init format

andrjohns:
#### Submisison Checklist

#### Summary:

Updates the recently added `log_prob` method so that the `constrained_params` argument expects its inputs to follow the same format/structure as the `init` argument for sampling/optimisation

#### Intended Effect:
Simpler usage of the `log_prob` method

#### How to Verify:
Additional tests included

#### Side Effects:
N/A

#### Copyright and Licensing


WardBrian:
Thanks for looking into this @andrjohns. I guess the downside to this approach is losing the ability to 'batch' calls?

andrjohns:
> Thanks for looking into this @andrjohns. I guess the downside to this approach is losing the ability to 'batch' calls?

Yeah pretty much. I was thinking an option would be to have multiple files, like how the inits are handled, but I'm not sure if that would be just as messy

src/cmdstan/command.hpp
```diff
@@ -644,50 +649,40 @@ int command(int argc, const char *argv[]) {
...
+                            return m_param_name == c_param_name;
+                          });
+
+        if (!present) {
+          msg << "Constrained value(s) for parameter " << m_param_name
+              << " not found!";
+          throw std::invalid_argument(msg.str());
+        }
+      }
+      model.transform_inits((*cpars_context), dummy_params_i, params_r_ind[0],
```
bob-carpenter, line 678: How do we know `params_r_ind` is non-empty so that indexing won't cause a segfault?

src/cmdstan/command.hpp
```diff
@@ -644,50 +649,40 @@ int command(int argc, const char *argv[]) {
...
-      Eigen::Map<Eigen::VectorXd, 0, StrideT> map_r(
-          c_params_r.data() + i, c_params_size, StrideT(1, c_params_vec_size));
-
-      stan::io::array_var_context context(param_names, map_r, param_dimss);
-      model.transform_inits(context, dummy_params_i, params_r_ind[i], &msg);
-    }
-
-    for (size_t i = c_params_vec_size; i < num_par_sets; i++) {
-      size_t iter = i - c_params_vec_size;
+    for (size_t i = cpars_size; i < num_par_sets; i++) {
```
bob-carpenter, line 684: generic loop increments should be `++i`.

src/cmdstan/arguments/arg_log_prob_constrained_params.hpp
```diff
@@ -6,9 +6,8 @@
 namespace cmdstan {
 /**
  * Argument for providing a file of parameters on the constrained scale
- * for use with the 'log_prob' method. The file can be in CSV or JSON format
- * and should contain a variable 'params_r' with either a vector or list/array
- * of vectors of constrained parameter values.
+ * for use with the 'log_prob' method. The file can be in JSON or R Dump format,
```
bob-carpenter, line 9: Do they need to be named with the `.json` and `.R` convention?  If so, please doc.  If not, please explain how it knows which is which (can just say "format is autodetected" or something if that's the case).

bob-carpenter (changes requested):
thanks for adding the doc strings for the classes.   just a few minor issues left that go beyond the doc and then this is good to go

src/cmdstan/command.hpp
andrjohns: The `params_r_ind` vector is initialised to the number of parameter sets passed in (both constrained and unconstrained), if no parameter sets are passed then the program throws before it gets to this indexing

andrjohns:
Sorry for the delay in getting to this @bob-carpenter! Ready for another look now

src/cmdstan/command.hpp
```diff
@@ -644,50 +649,34 @@ int command(int argc, const char *argv[]) {
...
-
-    if (c_params_size > 0 && c_params_size != param_names.size()) {
-      msg << "Incorrect number of constrained parameters provided! "
-             "Model has "
-          << param_names.size() << " parameters but " << c_params_size
-          << " were found.";
-      throw std::invalid_argument(msg.str());
-    }
-
+    int has_cpars = !cpars_file->is_default();
```
bob-carpenter, line 652: `has_cpars` should be declared as `bool`.  It will be cast it to an `int` implicitly if necessary elsewhere.

src/cmdstan/command.hpp
```diff
@@ -644,50 +649,34 @@ int command(int argc, const char *argv[]) {
...
     // Store in single nested array to allow single loop for calc and print
-    size_t num_par_sets = c_params_vec_size + u_params_vec_size;
+    size_t num_par_sets = u_params_vec_size + has_cpars;
```
bob-carpenter, line 654: This looks wrong because `has_cpars` will be either 0 (`false`) or ` (`true`).

src/cmdstan/command.hpp
```diff
@@ -644,50 +649,34 @@ int command(int argc, const char *argv[]) {
...
     std::vector<int> dummy_params_i;
-    for (size_t i = 0; i < c_params_vec_size; i++) {
-      Eigen::Map<Eigen::VectorXd, 0, StrideT> map_r(
-          c_params_r.data() + i, c_params_size, StrideT(1, c_params_vec_size));
+    if (!cpars_file->is_default()) {
+      std::string cpars_filename(cpars_file->value());
 
-      stan::io::array_var_context context(param_names, map_r, param_dimss);
-      model.transform_inits(context, dummy_params_i, params_r_ind[i], &msg);
+      std::shared_ptr<stan::io::var_context> cpars_context
```
bob-carpenter, line 660: [question]
Why is this getting turned into a pointer rather than just being declared as a `var_context`?

src/cmdstan/command.hpp
```diff
@@ -644,50 +649,34 @@ int command(int argc, const char *argv[]) {
...
+        }
+      }
+      model.transform_inits((*cpars_context), dummy_params_i, params_r_ind[0],
+                            &msg);
     }
 
-    for (size_t i = c_params_vec_size; i < num_par_sets; i++) {
-      size_t iter = i - c_params_vec_size;
+    // Use Map with inner stride to operate on all values from parameter set
+    using StrideT = Eigen::Stride<1, Eigen::Dynamic>;
```
bob-carpenter, line 677: [comment]
I don't understand what this is doing, so I hope it's right!

src/cmdstan/command.hpp
```diff
@@ -702,14 +691,14 @@ int command(int argc, const char *argv[]) {
 
     std::vector<std::string> p_names;
     model.constrained_param_names(p_names, false, false);
-    for (size_t i = 1; i < p_names.size(); i++) {
+    for (size_t i = 1; i < p_names.size(); ++i) {
       output_stream << "g_" << p_names[i] << ",";
     }
     output_stream << "g_" << p_names.back() << "\n";
```
bob-carpenter, line 697: [question]
Doesn't this skip the first entry and then output the last entry in `p_names` twice?  Is that intentional?  Some of this unexpected code could use some local doc.  For instance, why are we starting `i` at 1 rather than 0?

src/cmdstan/command.hpp
```diff
@@ -702,14 +691,14 @@ int command(int argc, const char *argv[]) {
...
     try {
       double lp;
       std::vector<double> gradients;
-      for (size_t i = 0; i < num_par_sets; i++) {
+      for (size_t i = 0; i < num_par_sets; ++i) {
```
bob-carpenter, line 701: Can't this just iterate over the elements of `params_r_ind` or is it not the same size as `num_par_sets`?  If not, then this could use some doc explaining the indexing.

src/test/test-models/bern_constrained_params_short.json
```diff
@@ -1 +1,4 @@
-{"params_r":[[0.5, -1.5, 2.6], [0.1, -3.5]]}
\ No newline at end of file
+{
```
bob-carpenter, line 1: There were many boundary conditions in that code and tests that wind up throwing.  I don't see any of that getting tested.  Are there already tests in place for all this?  If not, I think they should be added if possible.

bob-carpenter (dismissed):
I've never understood the design of CmdStan and find the code almost impossible to read as the boundary conditions all seems strange.

My main concern is that none of this seems to be tested, but then there may be tests elsewhere.

I'd be much more comfortable if you found someone other than me who actually understands the C++ in CmdStan to review.

src/test/test-models/bern_constrained_params_short.json
WardBrian: There are some tests from the first PR in `src/test/interface/log_prob_test.cpp`. They didn't get updated in this PR besides to change the input data for them it seems

src/cmdstan/command.hpp
andrjohns: This is referring to the number of "sets" of parameters being passed, and the method can only take one set of constrained parameters at a time

src/cmdstan/command.hpp
andrjohns: Because the existing function for reading in parameters/data (`get_var_context()`) returns an `std::shared_ptr`

andrjohns:
@WardBrian I've updated the test model to include various containers (to make that the un-constraining works correctly), and verified that the returned values are consistent with those returned by `cmdstanr`

bob-carpenter:
The doc needs work here.  

1.  The doc says
> The file can be in JSON or R Dump format, using the same structure as the 'init' argument.

What does this mean?  The init argument takes a specification for a single point, whereas it looks like this function is going to allow multiple points.  Something needs to specify what that format is.

2.  The doc needs to be super clear about what happens to the Jacobian adjustment.  Stan defines an unconstrained density by including the Jacobian for transforming from unconstrained back to constrained. This would naturally be `false` for providing log densities on the constrained scale, because that's the log density the user defines over constrained parameters in the model block.  On the other hand, if users are expecting this to match the `lp__` value that is output when the Jacobian flag is on, then that won't happen if we turn it off.

WardBrian:
If you’d like to not be the reviewer, you will still need to remove your existing “request changes”. We can then ask @rok-cesnovar to review?

bob-carpenter:
Thanks for letting me know.  I just dismissed my review.  I'd forgotten I looked at this a month ago and that suggestions submitted that way shouldn't be in the form of a review.

It does look like there are tests now, which is great. And the user-facing doc I think is important to clarify should go in the CmdStan docs, not here in the code.

rok-cesnovar (approved):
Looks good to me. Some user facing doc would be great, but that is outside of the scope of this repo.

Issue #1123, opened 2022-10-24
Laplace approximation sampler command


bob-carpenter:
#### Summary:

Add a method to perform sampling from a Laplace approximation.

#### Description:

Add a Laplace approximation sampler that returns a sample .csv file with:

* standard output configuration as comments at the top
* header for names of constrained parameters, `lp__` (target log density), and `lq__` (approximation log density)
* rows corresponding to draws on the constrained scale

The arguments required are

* `mode`:  a specification of a mode on the constrained scale using either
    * **`.csv` format**: a .csv file containing constrained parameters
        *  `#` *initial*: comments that will be ignored; these can be anywhere
        *  *header*: comma-separated names of constrained parameters in same format as our other output (e.g., sampling or optimization), with any names ending in double underscores (`__`) ignored; this must come before any value lines
        * *values*: there can be any number of rows for values, but only the final line is used
    * **`.json` format**:  json specification of unconstrained values in the same format as used for initialization in other methods; all parameter values must be specified, and non-parameter values will be ignored

* `jacobian`: boolean value indicating whether to include Jacobian adjustment or not;  including provides max a posteriori (MAP) estimate, not including produces a (penalized) maximum likelihood estimate (MLE); default is `true`

* `draws`: a positive integer indicating number of approximate draws to return; default is 100

* `refresh`: an integer indicating how often to provide messages signaling which iteration has been processed;  if this is not positive (<= 0), then no output will be produced for iterations (and ideally not for other messages from model class); default is 100

* `seed`: a short unsigned integer to use as the random number generator; default is to generate randomly based on time

#### Additional Information:

The service function implementing Laplace sampling is: https://github.com/stan-dev/stan/blob/develop/src/stan/services/optimize/laplace_sample.hpp

#### Current Version:
v2.30.1

PR #1134, opened 2022-11-25
cmdstan cleanup -  issues 1123, 1124, 1129, 1132, 1133

mitzimorris:
#### Submisison Checklist

#### Summary:

- Add method `laplace_sample`
- Add argument `jacobian` to method `optimize`; default is FALSE (current behavoir)
- For method `log_prob`, add functionality to allow sets of constrained parameters as inputs in the form of a Stan CSV file.
- Improve maintainability of CmdStan by splitting out helper functions into separate file `command_helper.hpp`.

Apologies for an omnibus PR - this is PR replaces #1128 - which unified treatment of argument `jacobian` for methods `laplace_sample` and `optimize` and also added file `command_helper.hpp`.   The methods in `command_helper.hpp` encapsulate the checking and marshalling of inputs for each method, with the goal of making the `command` function easier to maintain.

The logic for the `generate_quantities` method was broken out into a few helper functions, one of which gets the constrained parameters from the Stan CSV file.   This is exactly the functionality we need to send multiple sets of constrained parameters to method `log_prob`, so we refactored the logic for the log_prob method into a series of helper functions.

The function `command` still runs to 800+ non-comment lines.   Some code bloat could be fixed by further using the helper functions `get_arg_val` throughout and possibly refactoring logic for the `sample` and `variational` method.  Is it worth the effort here?  Unclear.

#### Intended Effect:

Add new features to CmdStan and improve readability and maintainability of code base.

#### How to Verify:

Unit tests

#### Documentation:

Documentation will be added to the CmdStan User's Guide in a separate PR on the docs repo.

#### Copyright and Licensing


src/cmdstan/command.hpp
```diff
@@ -860,7 +565,7 @@ int command(int argc, const char *argv[]) {
         info(
             "The number of warmup samples (num_warmup) must be greater than "
             "zero if adaptation is enabled.");
-        return_code = stan::services::error_codes::CONFIG;
+        return_code = return_codes::NOT_OK;
```
WardBrian, line 568: I agree that this is a good change for consistency, but I wonder if we should be moving toward trying to use more descriptive exit codes, rather than more generic ones

src/cmdstan/command.hpp
```diff
@@ -1284,15 +990,17 @@ int command(int argc, const char *argv[]) {
           logger, init_writers[0], sample_writers[0], diagnostic_writers[0]);
     }
   }
+  //////////////////////////////////////////////////
```
WardBrian, line 993: These comments - I think it useful to have something delineating the sections, but I'm not sure it is helpful if they're all the same. Maybe something like 

```cxx
// ------- VARIATIONAL START ------- //
code here;
// ------- VARIATIONAL END ------- //
```

src/cmdstan/command_helper.hpp
```diff
@@ -0,0 +1,587 @@
...
+  if (fname != "" && (stream.rdstate() & std::ifstream::failbit)) {
+    std::stringstream msg;
+    msg << "Can't open specified file, \"" << fname << "\"" << std::endl;
+    throw std::invalid_argument(msg.str());
+  }
+  return stream;
+}
+
+/**
+ * Parse a StanCSV output file created by the NUTS-HMC sampler and
```
WardBrian, line 264: I believe this comment is incorrect, there's no restriction making this specific to NUTS (I think anyway, this seems to be based on the code changed in https://github.com/stan-dev/cmdstan/pull/1106)

WardBrian (changes requested):
Thanks so much for taking this on. In addition to the new features, shaving ~300 lines off a file and reorganizing like you have done was much needed work.

I believe the core of the work here is good to go, see below for some lingering issues or questions I had

src/cmdstan/arguments/arg_optimize.hpp
```diff
@@ -14,9 +15,10 @@ class arg_optimize : public categorical_argument {
     _name = "optimize";
     _description = "Point estimation";
 
-    _subarguments.push_back(new arg_optimize_algo());
-    _subarguments.push_back(new arg_iter());
-    _subarguments.push_back(new arg_save_iterations());
+    _subarguments.emplace_back(std::move(new arg_optimize_algo()));
+    _subarguments.emplace_back(std::move(new arg_jacobian_false()));
+    _subarguments.emplace_back(std::move(new arg_iter()));
+    _subarguments.emplace_back(std::move(new arg_save_iterations()));
```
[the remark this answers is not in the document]
mitzimorris: I was following the implementation of `arg_log_prob.hpp`.
discussion on SO says that this doesn't necessarily matter, except when it does.  https://stackoverflow.com/questions/26860749/efficiency-of-c11-push-back-with-stdmove-versus-emplace-back-for-already

in this case, efficiency doesn't much matter,  so for the sake of consistency, we could use `push_back` everywhere.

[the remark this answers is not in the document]
WardBrian: Yeah, in that case I think I would also vote for `push_back` everywhere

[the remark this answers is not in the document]
mitzimorris: done

src/cmdstan/command.hpp
```diff
@@ -396,19 +200,24 @@ int command(int argc, const char *argv[]) {
...
+  // check filenames to avoid clobbering input files with output file
+  std::string input_fname;
+  std::string output_fname
+      = get_arg_val<string_argument>(parser, "output", "file");
+  if (user_method->arg("generate_quantities")) {
+    input_fname = get_arg_val<string_argument>(
+        parser, "method", "generate_quantities", "fitted_params");
+    if (input_fname.compare(output_fname) == 0) {
+      msg << "Filename conflict, fitted_params file " << input_fname
+          << " matches output filename, must be different." << std::endl;
```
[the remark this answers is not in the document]
mitzimorris: original message wasn't quite grammatical - I hope new version is both clear and concise.

src/cmdstan/command.hpp
```diff
@@ -430,6 +239,8 @@ int command(int argc, const char *argv[]) {
       = new_model(*var_context, random_seed, &std::cout);
 
   std::vector<std::string> model_compile_info = model.model_compile_info();
+  std::vector<std::string> model_params;
+  model.constrained_param_names(model_params, false, false);
```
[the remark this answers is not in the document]
mitzimorris: good catch!

src/cmdstan/command.hpp
```diff
@@ -517,300 +327,195 @@ int command(int argc, const char *argv[]) {
     parser.print(diagnostic_writers[i]);
   }
 
-  int refresh
-      = dynamic_cast<int_argument *>(parser.arg("output")->arg("refresh"))
-            ->value();
+  int refresh = get_arg_val<int_argument>(parser, "output", "refresh");
```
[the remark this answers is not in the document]
mitzimorris: indeed.   I could change it everywhere as part of the PR.

src/cmdstan/command.hpp
```diff
@@ -517,300 +327,195 @@ int command(int argc, const char *argv[]) {
...
-        = dynamic_cast<bool_argument *>(
-              parser.arg("method")->arg("optimize")->arg("save_iterations"))
-              ->value();
-
     if (algo->value() == "newton") {
-      return_code = stan::services::optimize::newton(
-          model, *(init_contexts[0]), random_seed, id, init_radius,
-          num_iterations, save_iterations, interrupt, logger, init_writers[0],
-          sample_writers[0]);
+      if (jacobian)
```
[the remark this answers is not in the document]
mitzimorris: seems reasonable.  will do.

src/cmdstan/command.hpp
mitzimorris: the stan::services layer is implemented with a set of posix return codes, defined here:  https://github.com/stan-dev/stan/blob/develop/src/stan/services/error_codes.hpp.  these aren't used consistently across the services, and, in fact, some of them are not used at all.

the definitions in std::services::error_codes are only really appropriate for C++ layers.   if the user is working in the shell, they don't really care about the return code, only the error message.   if the command is executed programmatically, either from a script, or Python, R, Julia, etc, then the error code needs to be interpreted correctly by the interface.

src/cmdstan/command.hpp
mitzimorris: good idea.  will do.

src/cmdstan/command_helper.hpp
WardBrian: It’s used mainly by `generate_quantities`, which supports non-NUTS inputs, so this is a bit confusing

mitzimorris: and it works?  that's interesting.  I'll check the "stan::io::read_csv" method - maybe that code is flexible enough to handle output files from other methods as well.   I'm not fond of that dependency, FWIW.

src/cmdstan/command.hpp
```diff
@@ -517,300 +327,195 @@ int command(int argc, const char *argv[]) {
...
+        return_code
+            = stan::services::optimize::newton<stan::model::model_base, true>(
+                model, *(init_contexts[0]), random_seed, id, init_radius,
+                num_iterations, save_iterations, interrupt, logger,
+                init_writers[0], sample_writers[0]);
+      else
+        return_code
+            = stan::services::optimize::newton<stan::model::model_base, false>(
```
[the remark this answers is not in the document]
mitzimorris: good idea.  can that be generalized for log_prob and laplace as well?

src/cmdstan/command_helper.hpp
WardBrian: Yeah, I made the (minimal) change in #1106 to support them. The only thing that it directly depended on was that HMC outputs 7 things before the user parameters, so now it just counts them.

src/cmdstan/command.hpp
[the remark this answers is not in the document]
WardBrian: I think so, yes. It wouldn’t make the repeated pattern any more concise, but at least shuffles it to one central place

src/cmdstan/command_helper.hpp
mitzimorris: cool!

src/cmdstan/command.hpp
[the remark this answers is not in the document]
mitzimorris: to generalized, how would I pass in the service method to the inline helper function?  as an argument or template parameter?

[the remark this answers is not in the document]
WardBrian: I don’t know enough about the variadic templating/arguments in C++. I *think* you can write something like 

```cxx
template <typename F, typename Ts…> auto jacobian_wrapper(F f, bool jacobian, Ts args…)
```

but I have no idea if this will work off the bat or if it is just easier to write a new function for each.

If it’s not possible as a template, it probably is as a macro, but I’d avoid that I think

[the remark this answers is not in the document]
bob-carpenter: Yes, you can use [parameter packs](https://en.cppreference.com/w/cpp/language/parameter_pack) to write variadic arguments recursively.  Here's a [tutorial with simple examples](https://kevinushey.github.io/blog/2016/01/27/introduction-to-c++-variadic-templates/).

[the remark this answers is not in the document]
mitzimorris: understand parameter packs - question is w/r/t function argument - I can just pass around function methods as args and invoke them?

```
if (jacobian) 
 return f<model, true>(Args&& args);
else
 return f<model, false>(Args&& args);
```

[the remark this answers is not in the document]
bob-carpenter: The expression`f<model, true>` doesn't make sense with `f` being an instance.   Every instance has to be concrete and specify all parameters.  So you have to make this decision higher up the call chain before `f` is instantiated or you have to be lazy and not instantiate `f` until you get inside the `if (jacobian)` conditional.   

I would suggest turning `jacobian` into a template parameter at a scope that includes `f`'s instantiation.  You can see that's what I did for the service functions for optimization.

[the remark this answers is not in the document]
WardBrian: I just tried for a while on godbolt and couldn't find a version which let me write down a template for `f`. 
https://godbolt.org/z/e6bje4ja6

It may still be possible, just my first attempt at writing it didn't work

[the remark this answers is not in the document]
bob-carpenter: You can't put template parameter on instances.  You can create a struct with a template and pass that in as follows.

```cpp
template <bool disable>
struct foo {
  int operator()(int x, double y) {
    return disable ? 0.0 : x * y;
  }
};

template <template <typename> class F, bool disable>
auto wrap_disable(Ts... xs) {
    F<disable> f;
    return f(xs...);
}
```

It pushes the construction of `F` done inside the `wrap_disable` function.  This will only work if the argument `F` can work with the single argument.

Maybe if you said what you're trying to do, we can work backward from there.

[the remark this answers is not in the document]
WardBrian: We have a bunch of places now where we have code that looks like

```cpp
if (jacobian)
    call_some_function<true>(a, bunch, of, args);
else
    call_some_function<false>(the,same,args,here);
```

My original inquiry was if there's a way to write a helper which handles this logic for us. However, I am beginning to suspect that the cure would be far worse than the original disease looking at the examples of code which does similar things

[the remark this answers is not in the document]
mitzimorris: this should work if that single template parameter is a parameter pack, no?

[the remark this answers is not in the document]
WardBrian: >  This will only work if the argument F can work with the single template parameter.

Yeah, I don't thing we'd want to turn the functions we're talking about here into structs (they're the services functions, mainly), _and_ some of them are called with `<model, false>` (e.g. more than one template parameter), while others are only called with the boolean for the jacobian adjustment.

I think I've seen enough to declare it's not worth trying to write a one-size-fits wrapper. That being said, we may want to consider writing (manually) some overloads for the service functions which wrap up this dispatch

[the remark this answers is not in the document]
mitzimorris: now I understand the problem.   no can do on this one.

src/cmdstan/command.hpp
```diff
@@ -489,8 +299,8 @@ int command(int argc, const char *argv[]) {
     auto output_filename = output_name + name_iterator(i) + output_ending;
     auto unique_fstream
         = std::make_unique<std::fstream>(output_filename, std::fstream::out);
-    if (!sig_figs_arg->is_default()) {
-      (*unique_fstream.get()) << std::setprecision(sig_figs_arg->value());
+    if (sig_figs > 0) {  // user-specified precision, default -1
```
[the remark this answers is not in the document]
mitzimorris: using "is_default", but doing it the stupid way.

src/cmdstan/arguments/arg_log_prob.hpp
```diff
@@ -3,7 +3,7 @@
 
 #include <cmdstan/arguments/arg_log_prob_unconstrained_params.hpp>
 #include <cmdstan/arguments/arg_log_prob_constrained_params.hpp>
-#include <cmdstan/arguments/arg_log_prob_jacobian.hpp>
+#include <cmdstan/arguments/arg_jacobian.hpp>
```
WardBrian, line 6: Nice catch on the duplication

src/cmdstan/command.hpp
```diff
@@ -163,7 +163,7 @@ int command(int argc, const char *argv[]) {
         list_argument *metric
             = dynamic_cast<list_argument *>(algo->arg("hmc")->arg("metric"));
         if (engine->value() != "nuts"
-            && (metric->value() != "dense_e" || metric->value() == "diag_e")) {
+            && !(metric->value() == "dense_e" || metric->value() == "diag_e")) {
```
WardBrian, line 166: Was this just a bug before? This change changes the logic of the conditional but the new version seems like what was actually intended all along?

WardBrian (approved):
LGTM. A couple small comments

I should note that for the new `laplace` code I can only really audit it to say that it looks reasonable and is tested, but I wouldn't say I'm familiar enough to confirm that it has all the arguments one might want for a laplace approximation, for example. I trust you on the design part there, but just wanted to say in case you wanted a second set of eyes on that

src/cmdstan/command.hpp
mitzimorris: I'm honestly not sure. 

To me, this looks like a kind of belt and suspsenders check.
Changing the logic didn't break the unit tests, which just means that this isn't actually tested.
There are two sampling algorithms:  hmc and fixed params, where "fixed_params" is a model without parameters,
and therefore no (or and empty) metric.

My call was to make the code agree with the comment, but that's being lazy.   The proper thing to do is clean up the logic, but I'm just not sure it's worth it.

mitzimorris:
The design was specified in issue https://github.com/stan-dev/cmdstan/issues/1123 - I believe that this implementation matches the spec.

src/cmdstan/command.hpp
bob-carpenter: First refactor the conditional by moving the operators inside the disjunction

```cpp
engine->value() != "nuts"
&& metric->value() != "dense_e"
&& metric->value() != "diag_e"
```

I don't think this is NUTS-specific, so I'd remove the first line.

Then the remaining conjuncts with

```cpp
metric->value() == "unit_e"
```

because that's the metric that does not support adaptation.

One might argue that the negated form is better because it explicitly negates the things that do have adaptation.  I'd also be OK with that reasoning, but I still think you can remove the `!= "nuts"` test.

WardBrian: > I still think you can remove the != "nuts" test.

The num_chains argument is also unsupported for `engine=static`, so somewhere it does need to check that it is NUTS

WardBrian (changes requested):
While confirming that the interface matches the spec (looks good!) I tried some intentionally broken commands and got some segmentation faults originating from the `mode` argument handling

src/cmdstan/arguments/arg_jacobian.hpp
```diff
@@ -8,11 +8,13 @@ namespace cmdstan {
...
-class arg_log_prob_jacobian : public bool_argument {
+class arg_jacobian : public bool_argument {
  public:
-  arg_log_prob_jacobian() : bool_argument() {
+  arg_jacobian() : bool_argument() {
     _name = "jacobian";
-    _description = "Apply jacobian adjustments to log-probability?";
+    _description
+        = "When true, include change-of-variables adjustment"
+          "for constraining parameter transforms";
```
WardBrian, line 17:
```suggestion
          " for constraining parameter transforms";
```

src/cmdstan/command_helper.hpp
```diff
@@ -0,0 +1,589 @@
...
+}
+
+/**
+ * Read fitted params from sample in StanCSV format.
+ *
+ * @param fname name of file which exists and has read perms.
+ * @param model Stan model
+ * @return Eigen vector of parameter estimates.
+ */
+std::vector<std::vector<double>> get_cparams_r_csv(
```
WardBrian, line 532: Since this is used in a call to log_prob_grad, the results here need to be put on to the unconstrained scale with `transform_inits` for them to be meaningful (c.f. the above comment and function)

src/cmdstan/command_helper.hpp
```diff
@@ -0,0 +1,589 @@
...
+  output_stream << std::setprecision(sig_figs) << "lp__,";
+  std::vector<std::string> p_names;
+  model.constrained_param_names(p_names, false, false);
+  for (size_t i = 0; i < p_names.size(); ++i) {
+    output_stream << "g_" << p_names[i];
+    if (i == p_names.size() - 1)
+      output_stream << "\n";
+    else
+      output_stream << ",";
+  }
```
WardBrian, line 567: This is actually a bug in the released version of this feature, but this needs to be `unconstrained_param_names`. 

Try it with a simplex parameter and you'll get a misformed CSV result:

```csv
lp__,g_theta.1,g_theta.2
-1.20397,-0.4
```

src/cmdstan/command_helper.hpp
```diff
@@ -0,0 +1,589 @@
...
+  Eigen::VectorXd theta_hat(param_names.size());
+  if (suffix(fname) == ".csv") {
+    get_theta_hat_csv(fname, param_names, theta_hat);
+  } else if (suffix(fname) == ".json") {
+    get_theta_hat_json(fname, param_names, theta_hat);
+  } else {
+    msg << "Mode file must be CSV or JSON, found " << fname << std::endl;
+    throw std::invalid_argument(msg.str());
+  }
+  return theta_hat;
```
WardBrian, line 438: Theta hat needs to be on the unconstrained scale (see non-review comment above or try a simplex model)

WardBrian (changes requested):
Okay, a few more things, all surrounding getting the right match up between unconstrained vs constrained representations

bob-carpenter:
> This to me implied that the parameters are not being unconstrained when they are read in, which is obviously a big problem. I assumed that must have been happening inside the services layer, but it seems like it should be happening here instead.

The service function in `stan-dev/stan` takes in unconstrained parameters.   There are multiple ways to create constrained inputs---either as JSON files or as vectors read out of output CSV files.  The interfaces all have ways to manipulate those so I thought it'd be easier to keep the service function for Laplace clean and leave the transforming to the interfaces.

src/cmdstan/command_helper.hpp
```diff
@@ -0,0 +1,589 @@
...
+/**
+ * Parse a StanCSV output file created by the optimizer and return a vector
+ * containing the estimates of the model parameters.
+ * Throws an exception if it cannot parse the CSV file.
+ *
+ * @param fname name of file which exists and has read perms.
+ * @param param_names vector of model constrained parameter names.
+ * @param theta_hat Eigen vector for estimated parameters
+ */
+void get_theta_hat_csv(const std::string &fname,
```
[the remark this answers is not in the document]
WardBrian: Another segfault: if the csv file provided has too few columns, the two four loops at the end will both index out of bounds

src/cmdstan/command.hpp
mitzimorris: refactored as helper function `validate_multi_chain_config`

working with the config for the sampler is a complete nightmare.  any access of a subarg which isn't actualy present will segfault.   to bullet-proof this would require more refactoring that it's worth, as there aren't that many folks using the command line directoy, and the CmdStanX wrapper interfaces vet the config beforehand.

bob-carpenter: Is there a way to test if a subarg exists?

mitzimorris: The `categorical_argument` differs from the `list_argument` in that it doesn't have an internal cursor which points into the list of subarguments.   I'm just not good enough at type/template hacking to see how to deal with this.
Bottom line is that dealing with the config of NUTS requires tedious traversal of the argument tree.

andrjohns:
Do you still need me to review any of this or has Brian covered everything?

mitzimorris:
I think Brian's got this covered.

src/cmdstan/command_helper.hpp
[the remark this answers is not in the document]
mitzimorris: added checks and unit tests accordingly.

src/cmdstan/command_helper.hpp
```diff
@@ -0,0 +1,589 @@
...
+/**
+ * Extract and validate constrained parameter inputs.
+ * Input file contains definitions for all model parmeters.
+ *
+ * @param fname name of file which exists and has read perms.
+ * @param model Stan model
+ * @return Eigen vector of parameter estimates.
+ */
+std::vector<std::vector<double>> get_cparams_r(
+    const std::string &fname, const stan::model::model_base &model) {
```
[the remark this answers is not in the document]
mitzimorris: correct - comment should be "Extract, validate, and unconstrain".
but logic which calls `transform_inits` is wrong - it only unconstrains the first row of inputs.
should be a loop.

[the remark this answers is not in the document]
mitzimorris: added a check on the unconstrain transform.

src/cmdstan/command_helper.hpp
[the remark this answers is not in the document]
mitzimorris: test for not optimization output:  https://github.com/stan-dev/cmdstan/blob/9a042823e83b59b26353b35afb2cc3e7c5eca988/src/test/interface/laplace_test.cpp#L82

test for truncated columns: https://github.com/stan-dev/cmdstan/blob/9a042823e83b59b26353b35afb2cc3e7c5eca988/src/test/interface/laplace_test.cpp#L114

mitzimorris:
> parameters are not being unconstrained when they are read in, which is obviously a big problem. I assumed that must have been happening inside the services layer, but it seems like it should be happening here instead.

yes, big problem and thanks for catching this one!

I fixed this and renamed a bunch of the command_helper functions.   I feel like it could or should be cleaner, but given that standalone_generated quantities does the unconstraining for you, and log_prob accepts constrained parameters as either JSON, Rdump  (single draw) or CSV (multiple draws), *and* laplace and generated quantities want Eigen::MatrixXd inputs while log_prob is simpler, I don't know how to make this cleaner, or if it's worth it.

src/cmdstan/command_helper.hpp
mitzimorris: refactored

src/cmdstan/command_helper.hpp
mitzimorris: fixed.

src/cmdstan/command_helper.hpp
mitzimorris: corrected, plus unit test to check.

mitzimorris:
ready for another round if you are!

WardBrian:
I can go through the changes first thing tomorrow. Excited to get this merged soon

src/cmdstan/command_helper.hpp
```diff
@@ -354,7 +448,18 @@ void get_theta_hat_csv(const std::string &fname,
...
   in.close();
-
+  // validate
+  if (!is_optimization) {
+    msg << "CSV file is not output from Stan optimization" << std::endl;
+    throw std::invalid_argument(msg.str());
+  }
+  // optimization output has col for "lp__" + model variables
+  if (names.size() < cparam_names.size()
+      || values.size() < cparam_names.size()) {
```
WardBrian, line 458: Is there a reason these are `<` and not `!=`? I guess they won't necessarily be the same length until after we've removed any of the `__` names, but it seems like it should also be a problem if they're too long.

I think if you move this block below where we calculate `col_offset`, we should instead check that `names.size() - col_offset == cparam_names.size()`, and that `values.size() == names.size()`.

This should fix the above segfault(s)

src/test/interface/laplace_test.cpp
```diff
@@ -137,3 +184,21 @@ TEST_F(CmdStan, laplace_jacobian_adjust) {
...
+  run_command_output out = run_command(cmd);
+  ASSERT_FALSE(out.hasError);
+
+  ss.str(std::string());
+  ss << convert_model_path(simplex_model) << " random seed=1234"
+     << " method=laplace mode=" << convert_model_path(simplex_mode_bad_csv)
+     << " 2>&1";
+  cmd = ss.str();
+  out = run_command(cmd);
+  ASSERT_TRUE(out.hasError);
```
WardBrian, line 203: Question: does `out.hasError` distinguish between a "nice" error (one we throw) and something like a segmentation fault?

src/cmdstan/command_helper.hpp
```diff
@@ -324,27 +358,87 @@ void get_fitted_params(const std::string &fname,
...
- * @param theta_hat Eigen vector for estimated parameters
+ * @param fname name of file which exists and has read perms
+ * @param model Stan model
+ * @return Eigen vector of unconstrained parameter estimates
  */
-void get_theta_hat_csv(const std::string &fname,
-                       const std::vector<std::string> &param_names,
-                       Eigen::VectorXd &theta_hat) {
+Eigen::VectorXd get_laplace_mode_csv(const std::string &fname,
+                                     const stan::model::model_base &model) {
```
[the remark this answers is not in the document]
WardBrian: Line 475 is the culprit.

src/cmdstan/command_helper.hpp
```diff
@@ -363,74 +468,75 @@ void get_theta_hat_csv(const std::string &fname,
...
+      cparams[i] = std::stod(values[i + col_offset]);
     } catch (const std::exception &e) {
-      msg << "Error parsing CSV file, bad value for " << param_names[i]
+      msg << "Error parsing CSV file, bad value for " << cparam_names[i]
           << std::endl;
       throw std::invalid_argument(msg.str());
     }
   }
+  std::vector<double> uparams = unconstrain_params(model, cparams);
+  return Eigen::Map<Eigen::VectorXd>(&uparams[0], uparams.size());
```
WardBrian, line 490: I'm not a C++ expert, is this implicitly returning a reference to a temporary value? It seems like something like this should lead to an invalid pointer later on, but maybe some memory magic is happening I don't understand.

WardBrian: My first google hit suggests this could indeed be a problem: https://stackoverflow.com/questions/38786856/returning-a-eigenmap-from-a-lambda - Eigen maps do not own their data

src/cmdstan/command_helper.hpp
```diff
@@ -354,7 +448,18 @@ void get_theta_hat_csv(const std::string &fname,
...
-
+  // validate
+  if (!is_optimization) {
+    msg << "CSV file is not output from Stan optimization" << std::endl;
+    throw std::invalid_argument(msg.str());
+  }
+  // optimization output has col for "lp__" + model variables
+  if (names.size() < cparam_names.size()
+      || values.size() < cparam_names.size()) {
+    msg << "CSV file is incomplete, expecting at least "
```
WardBrian, line 459: After the above change, we should also say "expecting exactly"

src/cmdstan/command_helper.hpp
```diff
@@ -363,74 +468,75 @@ void get_theta_hat_csv(const std::string &fname,
...
-      theta_hat[offset] = param_vec[i];
+  for (auto &&param_name : param_names) {
+    const auto param_vec = context->vals_r(param_name);
+    for (size_t i = 0; i < param_vec.size(); ++i) {
+      cparams[offset] = param_vec[i];
       ++offset;
     }
   }
+  std::vector<double> uparams = unconstrain_params(model, cparams);
+  return Eigen::Map<Eigen::VectorXd>(&uparams[0], uparams.size());
```
WardBrian, line 522: Same question as above

WardBrian (changes requested):
This looks really great. One more nagging set of CSV validity issues, and a couple questions on top of that

src/cmdstan/command_helper.hpp
mitzimorris: the optimization output may include transformed parameters and generated quantities values as well.
it needs to be at least as big as the number of constrained parameters, that's all we know.

WardBrian: Good point. I still think it is essential to take into account `col_offset` in these comparisons and to compare `values` and `names` (not `values` and `cparam_names`) to avoid the later indexing being unsafe

src/test/interface/laplace_test.cpp
mitzimorris: no, and it should.   fixing.

src/cmdstan/command_helper.hpp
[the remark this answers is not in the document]
mitzimorris: I cannot trigger this segfault.
haven't changed the code.

src/test/interface/laplace_test.cpp
mitzimorris: no it doesn't.   adding some logic to check for that.

src/cmdstan/command_helper.hpp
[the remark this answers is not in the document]
WardBrian: Well, UB isn’t guaranteed to segfault, but it’s pretty clear line 475 indexes out of bounds (in the currently pushed code) if `names.size() - cparam_names.size() < col_offset`.

The current if statements don’t catch this, since they only check that `names.size() >= cparam_names.size()`, so it won’t catch a model with 4 parameters being fed a CSV with lp_ and then 3 more columns after it, for example. This will then index out of bounds on `names`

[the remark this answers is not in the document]
mitzimorris: yes, managed to do this (hardcoded col offset to 10 - works like a charm - I just couldn't quite craft a CSV file that would make it blow up).  will tighten up the checks.

src/cmdstan/command_helper.hpp
mitzimorris: not using map - allocate and copy.   we're not coding any of this for efficiency

src/cmdstan/command_helper.hpp
```diff
@@ -468,7 +461,13 @@ Eigen::VectorXd get_laplace_mode_csv(const std::string &fname,
       break;
     }
   }
-
+  if (names.size() - col_offset < cparam_names.size()
+      || values.size() - col_offset < cparam_names.size()
+      || names.size() != values.size()) {
+    msg << "CSV file is incomplete, expecting at least "
+        << (cparam_names.size() + 1) << " values." << std::endl;
```
WardBrian, line 468: Confirmed the segfault I previously had is resolved and this now looks safe.

Two (minor) things:

Instead of "values" should we say columns? For example, if a column header is missing but the value is still there.

This message is a little weird specifically in the names != values case, since I do have at least 3 values, I just also have a fourth value

WardBrian:
Not sure if you are still tweaking in the background but based on what you've currently pushed this seems pretty much good to go to me

src/cmdstan/command_helper.hpp
mitzimorris: OK, will change message

mitzimorris:
> Not sure if you are still tweaking in the background but based on what you've currently pushed this seems pretty much good to go to me

still thinking about if it's possible to do further cleanup of the args config - another PR for that.

I will make the one change and then it's ready to merge.   thanks so much for the careful review!

WardBrian (approved):
LGTM!

Thanks for such a nice set of refactors and features.

mitzimorris:
more CSV tests for log_prob and laplace_sample

a. file only contains header comments
b. file contains config as header comment, column names row, but no data.

PR #1145, opened 2023-02-21
[CI] Ensure mac interface tests are running correctly on m1 instance

serban-nicusor-toptal:
#### Submisison Checklist

#### Summary:

WRT https://github.com/stan-dev/ci-scripts/issues/24 we've got a new M1 instance that we need to test our CI on to ensure everything is working fine.

#### How to Verify:

Results should be consistent with the older Mac Instance.

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company): Toptal

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -13,7 +13,7 @@ def setupCXX(CXX = env.CXX) {
 
 def runTests(String prefix = "") {
     """ make -j${env.PARALLEL} build
-      ${prefix}runCmdStanTests.py -j${env.PARALLEL} src/test/interface
+      python3 ${prefix}runCmdStanTests.py -j${env.PARALLEL} src/test/interface
     """
 }
 
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -173,70 +173,70 @@ pipeline {
                 }
             }
             parallel {
-                stage('Windows interface tests') {
-                    agent { label 'windows' }
-                    steps {
-                        setupCXX(WIN_CXX)
-                        runWinTests()
-                    }
-                    post {
-                        always {
+//                 stage('Windows interface tests') {
+//                     agent { label 'windows' }
+//                     steps {
+//                         setupCXX(WIN_CXX)
+//                         runWinTests()
+//                     }
+//                     post {
+//                         always {
+//
+//                             recordIssues id: "Windows",
+//                             name: "Windows interface tests",
+//                             enabledForFailure: true,
+//                             aggregatingResults : true,
+//                             tools: [
+//                                 gcc4(id: "Windows_gcc4", name: "Windows interface tests@GCC4"),
+//                                 clang(id: "Windows_clang", name: "Windows interface tests@CLANG")
+//                             ],
+//                             blameDisabled: false,
+//                             qualityGates: [[threshold: 1, type: 'TOTAL', unstable: true]],
+//                             healthy: 10, unhealthy: 100, minimumSeverity: 'HIGH',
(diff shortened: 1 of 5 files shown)


serban-nicusor-toptal:
The initial problem was the missing `python` binary on the M1 instance.
After that was fixed, we seem to have a failing test
```
test/interface/stansummary_test --gtest_output="xml:test/interface/stansummary_test.xml" failed
```

WardBrian:
Huh, it seems like the formatting/precision of the output numbers changed. They still look like the right values, just formatted differently. Odd behavior to see change like that

WardBrian:
Huh, this might have found an actual bug with how we compute the width and precision of `nan`

WardBrian:
@serban-nicusor-toptal I think I found the issue. Assuming it passes can do a run on all the machines to make sure it is consistent, but then I'll want someone (probably @rok-cesnovar) to take a quick look at the C++ changes I made

rok-cesnovar:
Happy to do it.

WardBrian:
Yep, that made it so CI matches my local (Ubuntu) results.

NaN values to the `calculate_width` function were previously returning `18446744071562067971`, which forced any column with an `nan` in it to be formatted as scientific notation. I think that the modern Clang library had a nicer implementation of the floating point standard which avoided this somehow. 

Explicitly treating `nan`s is the right thing to do, anyway.

serban-nicusor-toptal:
Just wanted to note that this [commit](https://github.com/stan-dev/cmdstan/pull/1145/commits/58c7b2cc5eed94c16c5d3927a63f01dbcdeac981) is still happening, I think that's the plugin itself doing the merge. If we change the default we might affect other projects on Jenkins hmmh

WardBrian:
Assuming the above run passes I think all we need to do is change the label from `m1` back to just `osx` and then merge.

@rok-cesnovar mind taking a look at the C++?

rok-cesnovar (approved):
C++ changes look good! Thanks both!

WardBrian:
@serban-nicusor-toptal is there a reason we have multiple `recordIssues` in this repo and only one of them in `stan`? It seems like all of them collect warnings from all the platforms

serban-nicusor-toptal:
The only reason for that is to have them visually separated, it makes it easier to debug and find targeted issues.
https://jenkins.flatironinstitute.org/job/Stan/job/CmdStan/job/develop/100/
vs
https://jenkins.flatironinstitute.org/job/Stan/job/Stan/job/develop/89/
Not sure why for Stan we went with a global one, maybe it's less to worry about ?
image

WardBrian:
It looks like (based on https://jenkins.flatironinstitute.org/job/Stan/job/CmdStan/job/PR-1145/12/) that the Windows and Linux gates were failing due to the warnings that the Mac was generating but ignoring

serban-nicusor-toptal:
Oh, that's unexpected, I was under the impression that it will treat them as different sources.
I think this is caused because of `aggregatingResults : true` 
``` 
aggregatingResults : boolean (optional)
By default, each static analysis result will be recorded as a separate result that is presented as an individual   
Jenkins Action with separate UI and dashboard. If you rather prefer aggregation of the results into a single result  
(i.e., single Jenkins Action), then activate this check box. You still can see the distribution of issues grouped by static analysis tool in the UI.
```
https://www.jenkins.io/doc/pipeline/steps/warnings-ng/
Which I think we want to be actually `false` in `cmdstan` case.

WardBrian:
Sounds good. When that passes I think you can go ahead and squash and merge

PR #1155, opened 2023-04-14
Feature/1149 pathfinder v2

mitzimorris:
#### Submisison Checklist

#### Summary:

adds logic to CmdStan's `command` method and argument parser to handle calls to Pathfinder.

#### Intended Effect:

expose Pathfinder.

#### How to Verify:

unit tests

#### Documentation:

will be done in a separate PR

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company):   Columbia University


mitzimorris:
Initial commit; adds logic for Pathfinder.
For optional diagnostic files, constructs `unique_stream_writer` and/or `json_writer` objects with nullptr to output stream.

Current branch `stan/multi-path` doesn't yet play nicely with json_writer.

mitzimorris:
this is ready for review.

src/test/interface/pathfinder_test.cpp
```diff
@@ -0,0 +1,114 @@
...
+  std::vector<std::string> test_result_single;
+  std::vector<std::string> test_result_diags;
+};
+
+TEST_F(CmdStan, pathfinder_good) {
+  std::stringstream ss;
+  ss << convert_model_path(multi_normal_model)
+     << " output refresh=0 file=" << convert_model_path(test_arg_output)
+     << " method=pathfinder num_psis_draws=40" run_command_output out
+      = run_command(ss.str());
```
WardBrian, line 36:
```suggestion
     << " method=pathfinder num_psis_draws=40";
      run_command_output out = run_command(ss.str());
```

src/cmdstan/arguments/arg_variational_iter.hpp
WardBrian, line 1: All these small lexical cast changes seem fine but may be better in a separate PR

src/cmdstan/command.hpp
```diff
@@ -125,32 +126,6 @@ int command(int argc, const char *argv[]) {
...
-  stan::math::init_threadpool_tbb(num_threads);
-
-  unsigned int num_chains = 1;
-  auto user_method = parser.arg("method");
-  if (user_method->arg("sample")) {
-    num_chains
-        = get_arg_val<int_argument>(parser, "method", "sample", "num_chains");
-    if (num_chains > 1)
-      validate_multi_chain_config(parser.arg("method"));
-  }
```
WardBrian, line 149: It appears that the `validate_multi_chain_config` function is no longer called anywhere. Either its functionality was replaced (in which case we can delete it from the helpers file) and I can't find it, or we lost this check somewhere and need to put it back

src/cmdstan/command.hpp
```diff
@@ -295,29 +211,84 @@ int command(int argc, const char *argv[]) {
...
+    int num_psis_draws
+        = get_arg_val<int_argument>(*pathfinder_arg, "num_psis_draws");
+    int num_paths = get_arg_val<int_argument>(*pathfinder_arg, "num_paths");
+    if (num_paths == 1) {
+      return_code = stan::services::pathfinder::pathfinder_lbfgs_single<
+          false, stan::model::model_base>(
+          model, *(init_contexts[0]), random_seed, id, init_radius,
+          history_size, init_alpha, tol_obj, tol_rel_obj, tol_grad,
+          tol_rel_grad, tol_param, max_lbfgs_iters, num_elbo_draws, num_draws,
+          false, refresh, interrupt, logger, init_writer, sample_writers[0],
```
WardBrian, line 261: Is there a reason why the `save_iterations` argument isn't used (in place of the `false` on this line)?

src/cmdstan/command.hpp
```diff
@@ -295,29 +211,84 @@ int command(int argc, const char *argv[]) {
...
+      write_stan(pathfinder_writer);
+      write_model(pathfinder_writer, model.model_name());
+      write_datetime(pathfinder_writer);
+      parser.print(pathfinder_writer);
+      return_code = stan::services::pathfinder::pathfinder_lbfgs_multi<
+          stan::model::model_base>(
+          model, init_contexts, random_seed, id, init_radius, history_size,
+          init_alpha, tol_obj, tol_rel_obj, tol_grad, tol_rel_grad, tol_param,
+          max_lbfgs_iters, num_elbo_draws, num_draws, num_psis_draws, num_paths,
+          true, refresh, interrupt, logger, init_writers, sample_writers,
```
WardBrian, line 286: Similarly, `save_iterations` instead of hard-coded `true`?

src/cmdstan/command.hpp
```diff
@@ -295,29 +211,84 @@ int command(int argc, const char *argv[]) {
...
+          diagnostic_json_writers, pathfinder_writer,
+          diagnostic_json_writers[0]);
```
WardBrian, line 288: Is this passing the same JSON diagnostic writer as both the writer for the first pathfinder and for the overall multi-pathfinder?

src/test/interface/pathfinder_test.cpp
```diff
@@ -0,0 +1,114 @@
...
+  std::fstream result_stream(convert_model_path(test_result_8path));
+  std::stringstream result_sstream;
+  result_sstream << result_stream.rdbuf();
+  result_stream.close();
+  std::string output = result_sstream.str();
+  EXPECT_EQ(1, count_matches("# Elapsed Time:", output));
+  EXPECT_EQ(1, count_matches(" seconds (Pathfinder)", output));
+}
+
+TEST_F(CmdStan, pathfinder_diagnostic_json) {
```
WardBrian, line 97: Can you also add a test for multi-path for both kinds of diagnostic file?

src/test/interface/pathfinder_test.cpp
WardBrian, line 1: There are several sub-arguments to pathfinder that are not tested by this file (e.g. `save_iterations`, `num_draws`, etc) that would be good to have some basic tests for

WardBrian (changes requested):

src/cmdstan/command.hpp
mitzimorris: good catch.  we should have this - moved too much stuff around.  added back.

src/cmdstan/command.hpp
mitzimorris: no.  fixed.

src/cmdstan/command.hpp
WardBrian: I noticed another function in `helper` called `allow_multichain` which seems to do mostly the same thing (+ some pathfinder logic), but doesn't look exactly the same

src/cmdstan/command.hpp
mitzimorris: no.   diagnostic writers are only on the individual single-path pathfinders.  the overall multi-pathfinder writer is the importance resampled draws from the single-path pathfinders.  no gradients involved.

src/test/interface/pathfinder_test.cpp
mitzimorris: there's only one kind of diagnostic file.

src/cmdstan/command.hpp
mitzimorris: that's used by another helper function - `get_num_chains`.  added back `validate_multi_chain_config` call, also using `get_num_chains` since with pathfinder,  `num_paths` also determines `num_chains`.

src/cmdstan/command.hpp
mitzimorris: added dummy_writer.

src/test/interface/pathfinder_test.cpp
mitzimorris: added tests for max_lbfgs_iters and num_draws.

src/test/interface/pathfinder_test.cpp
```diff
@@ -0,0 +1,160 @@
...
+
+  std::fstream result_stream(convert_model_path(test_result_diags));
+  std::stringstream result_sstream;
+  result_sstream << result_stream.rdbuf();
+  result_stream.close();
+  std::string output = result_sstream.str();
+  ASSERT_FALSE(output.empty());
+  rapidjson::Document document;
+  ASSERT_FALSE(document.Parse<0>(output.c_str()).HasParseError());
+  EXPECT_EQ(1, count_matches("\"10\" : {\"iter\" : 10", output));
```
WardBrian, line 139: This appears to be failing

src/cmdstan/command_helper.hpp
```diff
@@ -634,6 +670,190 @@ void services_log_prob_grad(const stan::model::model_base &model, bool jacobian,
...
+  bool adapt_engaged
+      = dynamic_cast<bool_argument *>(adapt->arg("engaged"))->value();
+  list_argument *algo
+      = dynamic_cast<list_argument *>(sample_arg->arg("algorithm"));
+  if (algo->value() == "hmc") {
+    list_argument *engine
+        = dynamic_cast<list_argument *>(algo->arg("hmc")->arg("engine"));
+    list_argument *metric
+        = dynamic_cast<list_argument *>(algo->arg("hmc")->arg("metric"));
+    if (adapt_engaged && engine->value() == "nuts" && metric->value() != "unit")
```
WardBrian, line 698: I don't believe "unit" is a valid metric - "unit_e"?

src/cmdstan/command_helper.hpp
```diff
@@ -634,6 +670,190 @@ void services_log_prob_grad(const stan::model::model_base &model, bool jacobian,
...
+  categorical_argument *adapt
+      = dynamic_cast<categorical_argument *>(sample_arg->arg("adapt"));
+  bool adapt_engaged
+      = dynamic_cast<bool_argument *>(adapt->arg("engaged"))->value();
+  list_argument *algo
+      = dynamic_cast<list_argument *>(sample_arg->arg("algorithm"));
+  if (algo->value() == "hmc") {
+    list_argument *engine
+        = dynamic_cast<list_argument *>(algo->arg("hmc")->arg("engine"));
+    list_argument *metric
```
WardBrian, line 696: The second half of this function is repeating the logic of the `validate_multi_chain_config` function but just returns a boolean rather than raising an error. We should only be doing this one place

src/cmdstan/command_helper.hpp
```diff
@@ -634,6 +670,190 @@ void services_log_prob_grad(const stan::model::model_base &model, bool jacobian,
...
+    return get_arg_val<int_argument>(parser, "method", "pathfinder",
+                                     "num_paths");
+  if (!user_method->arg("sample"))
+    return 1;
+  unsigned int num_chains
+      = get_arg_val<int_argument>(parser, "method", "sample", "num_chains");
+  if (num_chains > 1 && !allow_multichain(parser))
+    throw std::invalid_argument(
+        "Argument 'num_chains' can currently only be used for NUTS with "
+        "adaptation and dense_e or diag_e metric");
```
WardBrian, line 726: This is also duplicate code when `validate_multi_chain_config` is considered.

I think most likely we should just remove `allow_multichain` and this validation and use the pre-existing `validate_multi_chain_config` if we're sampling

src/cmdstan/command.hpp
```diff
@@ -295,29 +222,79 @@ int command(int argc, const char *argv[]) {
...
+    double tol_rel_obj
+        = get_arg_val<real_argument>(*pathfinder_arg, "tol_rel_obj");
+    double tol_grad = get_arg_val<real_argument>(*pathfinder_arg, "tol_grad");
+    double tol_rel_grad
+        = get_arg_val<real_argument>(*pathfinder_arg, "tol_rel_grad");
+    double tol_param = get_arg_val<real_argument>(*pathfinder_arg, "tol_param");
+    int max_lbfgs_iters
+        = get_arg_val<int_argument>(*pathfinder_arg, "max_lbfgs_iters");
+    bool save_iterations
+        = get_arg_val<bool_argument>(*pathfinder_arg, "save_iterations");
```
WardBrian, line 255: This flag is still not used anywhere in the following code, which has hardcoded true/false values for the same-named service function argument

WardBrian: This needs to be tested

src/cmdstan/command.hpp
```diff
@@ -139,154 +154,66 @@ int command(int argc, const char *argv[]) {
...
-
-  std::vector<stan::callbacks::unique_stream_writer<std::ostream>>
-      sample_writers;
-  sample_writers.reserve(num_chains);
-  std::vector<stan::callbacks::unique_stream_writer<std::ostream>>
-      diagnostic_writers;
-  diagnostic_writers.reserve(num_chains);
+  // Setup callbacks
+  stan::callbacks::interrupt interrupt;
+  stan::callbacks::writer dummy_writer;
```
WardBrian, line 191: Because it's currently unusued it does not matter, but I suspect if it ever becomes used the Multi-path pathfinder service will expect a structured writer here rather than a `writer`

WardBrian (changes requested):

src/cmdstan/command_helper.hpp
mitzimorris: good catch - fixed!

src/cmdstan/command_helper.hpp
mitzimorris: much cleaner - thanks.

src/cmdstan/command.hpp
mitzimorris: right.  testing this, and there's  definitely a problem here.  seems to be always saving iters, no matter what the setting is.

WardBrian: As currently written it should always save iterations for multi-path pathfinder and never save them for single path. This is assuming it is all correct in the services layer, of course

mitzimorris: this is the problem:   "save_iterations" means dump lbfgs information to diagnostic_writer.  setting "save_iterations=1" (true), only makes sense if the output sub-arg "diagnostic_file" is specified.  we shouldn't expose "save_iterations" at all; instead, set the bool value "save_iterations" to whether or not arg "diagnostic_file" is specified.

mitzimorris: removed "save_iterations" as user-facing argument.  feature tested via tests for diagnostic_file.

src/cmdstan/command.hpp
mitzimorris: changed to json_writer.

src/test/interface/pathfinder_test.cpp
mitzimorris: tweaked test logic - should reliably pass.

mitzimorris:
changes made - one more time?

WardBrian (commented):
A few minor questions/style comments but otherwise I think this is looking pretty good.

While playing around I noticed that the CSV for multi-pathfinder appears to be incorrect (the header has the columns in the wrong order). So that needs to be fixed before we merge this (ping @SteveBronder)

src/cmdstan/command.hpp
```diff
@@ -295,29 +218,82 @@ int command(int argc, const char *argv[]) {
...
+    } else {
+      std::string output_file
+          = get_arg_val<string_argument>(parser, "output", "file");
+      std::string pf_name;
+      std::string pf_suffix;
+      get_basename_suffix(output_file, pf_name, pf_suffix);
+      if (pf_suffix.empty())
+        pf_suffix = ".csv";
+      auto ofs = std::make_unique<std::ofstream>(pf_name + "_pathfinder"
+                                                 + pf_suffix);
```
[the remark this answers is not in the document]
mitzimorris: evidently - posteriordb has some csv files that have suffix ".tmp".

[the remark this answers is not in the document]
WardBrian: Today I learned. Sounds reasonable.

Was there a discussion anywhere of the `_pathfinder` naming convention?

src/cmdstan/command_helper.hpp
```diff
@@ -635,6 +672,158 @@ void services_log_prob_grad(const stan::model::model_base &model, bool jacobian,
...
+        stan::callbacks::json_writer<std::ofstream> jwriter(std::move(ofs));
+        diag_json_writers.emplace_back(std::move(jwriter));
+      }
+    }
+  } else {
+    diag_csv_writers.reserve(num_chains);
+    if (diagnostic_file.empty()) {
+      for (int i = 0; i < num_chains; ++i) {
+        diag_csv_writers.emplace_back(nullptr, "# ");
+      }
```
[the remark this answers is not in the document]
mitzimorris: when running pathfinder, create vector of csv writers with nullptrs and when running sampler create vector of json writers with nullptrs?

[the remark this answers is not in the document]
WardBrian: I think that is the simplest thing, and I believe it will only take 2*num_chains bytes worth of extra space compared to what is being done now

src/cmdstan/command.hpp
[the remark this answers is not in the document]
mitzimorris: @SteveBronder

> Was there a discussion anywhere of the _pathfinder naming convention?

WardBrian:
Another question. Right now, if one runs ADVI, you get the following as part of the output:

```

------------------------------------------------------------
EXPERIMENTAL ALGORITHM:
  This procedure has not been thoroughly tested and may be unstable
  or buggy. The interface is subject to change.
------------------------------------------------------------
```

Is there a desire or need to have a similar warning here?

mitzimorris:
I would hope not.  have we tested this pathfinder algorithm on PosteriorDB and does it get the same (or similar) results as the R implementation evaluations reported in the Pathfinder paper?   @SteveBronder ?

WardBrian:
I was primarily asking because of @bob-carpenter's remark in the original issue (https://github.com/stan-dev/stan/issues/3119) 
>This is all going in the experimental folder and will be marked as such like ADVI when run.

mitzimorris:
yes, I noticed the remark in the initial issue.  personally, I would prefer one less directory in the path, but that's just to save keystrokes.

mitzimorris:
this failed because downstream checks had warnings from boost::lexical_cast using "sprintf".   the fix for this is https://github.com/stan-dev/math/pull/2919 is someone wants to review and approve.

WardBrian:
I wonder why other PRs don’t seem to fail because of that warning. I don’t think it’s a blocker for merging this at any rate

src/cmdstan/command_helper.hpp
```diff
@@ -785,17 +785,24 @@ void init_callbacks(
...
-      for (int i = 0; i < num_chains; ++i) {
-        diag_json_writers.emplace_back(
-            stan::callbacks::json_writer<std::ofstream>());
-      }
-    } else {
-      std::vector<std::string> diag_filenames;
+  if (!diagnostic_file.empty()) {
+    std::vector<std::string> diag_filenames;
+    if (user_method->arg("pathfinder")) {
+      diag_json_writers.resize(0);
```
WardBrian, line 805: Basic C++ question: is there a reason to prefer `.resize(0)` to `.clear()`?

mitzimorris: maybe?  discussion inconclusive:  https://stackoverflow.com/questions/2738967/vector-clear-vs-resize

WardBrian: It sounds like in C++11 and later clear is not allowed to change the capacity, so either seems fine.

WardBrian:
I think the only remaining question I have before approval is whether or not we believe we should be calling `stan::services::util::experimental_message` to produce a warning like ADVI

Thoughts @SteveBronder @bob-carpenter?

src/cmdstan/command_helper.hpp
WardBrian: I believe the current test failures are actually caused by the `.remove(0)`, since `remove` requires default-constructable types. Changing it to clear() resolves the issue for me locally.

WardBrian:
@mitzimorris you can fix the CI failure by changing `resize(0)` to `clear()` on those std::vectors. The reason linking is failing is that the unique_stream_writer has a forward-declared default constructor, but it never gets defined (this is probably a separate issue @SteveBronder should fix, but `clear()` will work and is more semantically meaningful)

src/cmdstan/command.hpp
```diff
@@ -294,29 +217,80 @@ int command(int argc, const char *argv[]) {
...
+          history_size, init_alpha, tol_obj, tol_rel_obj, tol_grad,
+          tol_rel_grad, tol_param, max_lbfgs_iters, num_elbo_draws, num_draws,
+          save_iterations, refresh, interrupt, logger, init_writer,
+          sample_writers[0], diagnostic_json_writers[0]);
+    } else {
+      std::string output_file
+          = get_arg_val<string_argument>(parser, "output", "file");
+      std::string pf_name;
+      std::string pf_suffix;
+      get_basename_suffix(output_file, pf_name, pf_suffix);
```
SteveBronder, line 270: Instead of functions that have input/output arguments can we have this return a `std::pair<std::string, std::string>`?

src/cmdstan/command.hpp
```diff
@@ -294,29 +217,80 @@ int command(int argc, const char *argv[]) {
...
+      if (pf_suffix.empty())
+        pf_suffix = ".csv";
+      auto ofs = std::make_unique<std::ofstream>(pf_name + "_pathfinder"
+                                                 + pf_suffix);
+      if (sig_figs > -1)
+        ofs->precision(sig_figs);
```
SteveBronder, line 276:
```suggestion
      if (sig_figs > -1) {
        ofs->precision(sig_figs);
     }
```

src/cmdstan/command_helper.hpp
```diff
@@ -635,6 +672,158 @@ void services_log_prob_grad(const stan::model::model_base &model, bool jacobian,
...
+ * @param parser user config
+ * @return int num chains or paths
+ */
+unsigned int get_num_chains(argument_parser &parser) {
+  auto user_method = parser.arg("method");
+  if (user_method->arg("pathfinder"))
+    return get_arg_val<int_argument>(parser, "method", "pathfinder",
+                                     "num_paths");
+  if (!user_method->arg("sample"))
+    return 1;
```
SteveBronder, line 691: I prefer curlies around the bodies of its. It mostly helps readability and to make sure someone in the future doesn't think they are in an if when it only effects the next line

src/cmdstan/command_helper.hpp
```diff
@@ -635,6 +672,158 @@ void services_log_prob_grad(const stan::model::model_base &model, bool jacobian,
...
+    if (input_file.empty()) {
+      throw std::invalid_argument(
+          std::string("Argument fitted_params file - found empty string, "
+                      "expecting filename."));
+      if (input_file.compare(sample_file) == 0) {
+        std::stringstream msg;
+        msg << "Filename conflict, fitted_params file " << input_file
+            << " and output file names are identical, must be different."
+            << std::endl;
+        throw std::invalid_argument(msg.str());
```
SteveBronder, line 722: Can you just use an std string here?

src/cmdstan/command_helper.hpp
```diff
@@ -635,6 +672,158 @@ void services_log_prob_grad(const stan::model::model_base &model, bool jacobian,
...
+/**
+ * Check possible name conflicts between input and output files where both
+ * are in Stan CSV format; if conflicts detected, quit with an error message.
+ *
+ * @param parser user config
+ */
+void check_file_config(argument_parser &parser) {
+  std::string sample_file
+      = get_arg_val<string_argument>(parser, "output", "file");
+  std::string input_file;
```
SteveBronder, line 708: Instead of defining `input_file` here can we just define it where it's constructed in the if statement bodies?

src/cmdstan/command_helper.hpp
```diff
@@ -635,6 +672,158 @@ void services_log_prob_grad(const stan::model::model_base &model, bool jacobian,
...
+        &diag_json_writers) {
+  auto user_method = parser.arg("method");
+  unsigned int num_chains = get_num_chains(parser);
+  unsigned int id = get_arg_val<int_argument>(parser, "id");
+  int sig_figs = get_arg_val<int_argument>(parser, "output", "sig_figs");
+
+  sample_writers.reserve(num_chains);
+  std::vector<std::string> output_filenames;
+  make_filenames(get_arg_val<string_argument>(parser, "output", "file"), ".csv",
+                 num_chains, id, output_filenames);
```
SteveBronder, line 781: Can we rewrite `make_filenames` to return an std vector of strings?

src/cmdstan/command_helper.hpp
```diff
@@ -635,6 +672,158 @@ void services_log_prob_grad(const stan::model::model_base &model, bool jacobian,
...
+}
+
+void init_callbacks(
+    argument_parser &parser,
+    std::vector<stan::callbacks::unique_stream_writer<std::ofstream>>
+        &sample_writers,
+    std::vector<stan::callbacks::unique_stream_writer<std::ofstream>>
+        &diag_csv_writers,
+    std::vector<stan::callbacks::json_writer<std::ofstream>>
+        &diag_json_writers) {
```
SteveBronder, line 772: Are all 3 of these vectors input/output arguments? If they come in as uninitialized arguments then I'd rather return a tuple or struct holding the different writers rather than have them be input / output arguments

SteveBronder (changes requested):
In general looks good, though I'd like to avoid using input/output arguments so much and just pairs/tuples/structs with the objects constructed in the function

src/cmdstan/command_helper.hpp
mitzimorris: in this case, I disagree because and if which wraps a return stmt w/out curly braces is pretty unambiguous.

src/cmdstan/command_helper.hpp
mitzimorris: I think the filename is useful information - is that what you're asking?  or are you suggesting avoid setting up the std::stringstream and just constructing the message as argument to the `std::invalid_argument` constructor?

I would prefer to leave it as is because
+  this makes it easy to see the message
+ it's consistent with error messaging elsewhere.

src/cmdstan/command_helper.hpp
mitzimorris: yes, they come in as uninitialized args, but creating a tuple struct doesn't make much sense here, given the calling structure for the services.  the main goal of `init_callbacks` is to move a chunk of tedious code out of the main part of the `command` function, in order to make it a little easier to see the overall logic flow.
not changing.

src/cmdstan/command_helper.hpp
mitzimorris: fixed!

mitzimorris:
requested changes made.  ready for re-re-re-review.

WardBrian (approved):
I’m happy with the changes, I’ll leave final review for @SteveBronder

src/cmdstan/command_helper.hpp
```diff
@@ -635,6 +675,157 @@ void services_log_prob_grad(const stan::model::model_base &model, bool jacobian,
...
```
SteveBronder, line 774: (optional) I think it could be nice to have a little struct for these that could be like

```c++
struct Callbacks {
  std::vector<stan::callbacks::unique_stream_writer<std::ofstream>> sample_csv;
  std::vector<stan::callbacks::unique_stream_writer<std::ofstream>> diagnostic_csv;
  std::vector<stan::callbacks::json_writer<std::ofstream>> diagnostic_json;
  explicit Callbacks(argument_parser& parser) {
    auto user_method = parser.arg("method");
    unsigned int num_chains = get_num_chains(parser);
    unsigned int id = get_arg_val<int_argument>(parser, "id");
    int sig_figs = get_arg_val<int_argument>(parser, "output", "sig_figs");

    sample_writers.reserve(num_chains);
    std::vector<std::string> output_filenames
        = make_filenames(get_arg_val<string_argument>(parser, "output", "file"),
                        ".csv", num_chains, id);
    for (int i = 0; i < num_chains; ++i) {
      auto ofs = std::make_unique<std::ofstream>(output_filenames[i]);
      if (sig_figs > -1)
        ofs->precision(sig_figs);
      sample_writers.emplace_back(std::move(ofs), "# ");
    }

    diag_json_writers.reserve(num_chains);
    diag_csv_writers.reserve(num_chains);
    // create no-op writers by default
    for (int i = 0; i < num_chains; ++i) {
      diag_json_writers.emplace_back(
          stan::callbacks::json_writer<std::ofstream>());
    }
    for (int i = 0; i < num_chains; ++i) {
      diag_csv_writers.emplace_back(nullptr, "# ");
    }
    // create json, csv writers as needed.
    std::string diagnostic_file
        = get_arg_val<string_argument>(parser, "output", "diagnostic_file");
    if (!diagnostic_file.empty()) {
      std::vector<std::string> diag_filenames;
      if (user_method->arg("pathfinder")) {
        diag_json_writers.clear();
        diag_filenames = make_filenames(diagnostic_file, ".json", num_chains, id);
        for (int i = 0; i < num_chains; ++i) {
          auto ofs = std::make_unique<std::ofstream>(diag_filenames[i]);
          if (sig_figs > -1)
            ofs->precision(sig_figs);
          stan::callbacks::json_writer<std::ofstream> jwriter(std::move(ofs));
          diag_json_writers.emplace_back(std::move(jwriter));
        }
      } else {
        diag_csv_writers.clear();
        diag_filenames = make_filenames(diagnostic_file, ".csv", num_chains, id);
        for (int i = 0; i < num_chains; ++i) {
          auto ofs = std::make_unique<std::ofstream>(diag_filenames[i]);
          if (sig_figs > -1)
            ofs->precision(sig_figs);
          diag_csv_writers.emplace_back(std::move(ofs), "# ");
        }
      }
    }
  }
};
````

Because then we call set this up like 

```c++
Callbacks writers(parser);
```

SteveBronder (approved):
One optional comment but lgmt!!

mitzimorris:
current implementation of Pathfinder has default num_draws set to 2000.  why? @bob-carpenter - opinion?
NUTS-HMC runs 2000 iterations by default becuase there are 1000 warmup iters and 1000 sampling iters.

bob-carpenter:
Our default for HMC produces 1000 warmup iterations and 1000 draws.  That's why I suggested 1000 draws as the default number of Pathfinder draws.   1024 draws from the posterior would bring MCMC standard error down to about std-deviation / 32.  But we're not even getting true posterior draws here and the importance ratios is a much better indicator of error than sample size.

SteveBronder:
I think we just need to fix the merge conflict and we are good to go!

mitzimorris:
@syclik @WardBrian, @serban-nicusor-toptal - the issue here is boost::lexical_cast.   proposed solution - update version of boost library?

WardBrian:
My proposed solution for this PR is just to merge - we know about the warning, and everything else passed as-is

Re, updating boost: https://github.com/stan-dev/math/issues/2856

Issue #1180, opened 2023-08-14
multi-pathfinder algorithm - add arg to enable/supress output of single-path pathfinders


mitzimorris:
#### Summary:
For multi-path Pathfinder, add a command-line argument to enable/supress writing of CSV files for the individual single-path pathfinders.

#### Description:
When running multi-path pathfinder, the algorithm outputs the draws from each single-path pathfinder as well as the PSIS draws.  The single-path CSV files are there mainly for the purpose of checking the Pathfinder algorithm; the PSIS draws are the outputs intended for inference.

#### Current Version:
v2.32.2

WardBrian:
My vote is disabled by default with a flag to enable

mitzimorris:
there's this argument grouping - adding this to "output" seems most natural, but this is totally pathfinder-specific.   any suggestions for arg name?

WardBrian:
The assumption that output spec is independent of algorithm looks more questionable by the day. I think it really only makes sense to place under pathfinder, since putting it under output would add an argument to all the other methods which don't need it. 

I think calling it `save_individual_paths` is good. The edge case of `num_paths=1 save_individual_paths=0` is a bit annoying, but I think we can just ignore it entirely if num_paths is 1

bob-carpenter:
This was, of course, the failure mode I anticipated for this design, so I'm also OK putting it under the pathfinder hierarchy of command-line args.

mitzimorris:
proposed fix:

pathfinder boolean arg `save_single_paths`, default false.

the PSIS sample file is `output.csv`

by default, pathfinder method runs 4 single-path pathfinders.
when  `save_single_paths`, single-path pathfinder output filenames look like:
`output_pathfinder_<path-id>.csv`, e.g.

`output_pathfinder_1.csv` ... `output_pathfinder_4.csv`

if arg `num_paths` == 1,  the result of the single-path pathfinder run will also be in file `output.csv`.
reasoning here is that if you only get one output file, you don't need any qualifiers on the filename.

happy to modify this as @bob-carpenter and @WardBrian see fit.

bob-carpenter:
That sounds OK to me.

mitzimorris:
single-path output name suggestions?

`output_pathfinder_1.csv` vs `output_single_path_1.csv`. ???

WardBrian:
I would say output_path_1

PR #1184, opened 2023-08-17
`Feature/1180 pathfinder single path csvs

mitzimorris:
#### Submisison Checklist

#### Summary:

Added flag `save_single_path` to method `pathfinder`, per discussion in #1180 

#### Intended Effect:

Better user experience.

#### How to Verify:

Unit tests

#### Side Effects:

N/A

#### Documentation:

see PR https://github.com/stan-dev/docs/pull/651#

#### Copyright and Licensing

### DIFF
--- a/src/cmdstan/command.hpp
+++ b/src/cmdstan/command.hpp
@@ -195,8 +195,10 @@ int command(int argc, const char *argv[]) {
       diagnostic_csv_writers;
   std::vector<stan::callbacks::json_writer<std::ofstream>>
       diagnostic_json_writers;
+  stan::callbacks::unique_stream_writer<std::ofstream> pathfinder_writer;
+
   init_callbacks(parser, sample_writers, diagnostic_csv_writers,
-                 diagnostic_json_writers);
+                 diagnostic_json_writers, pathfinder_writer);
 
   // Setup initial parameter values - arg "init"
   // arg is either filename or init radius value
@@ -264,18 +266,6 @@ int command(int argc, const char *argv[]) {
           save_iterations, refresh, interrupt, logger, init_writer,
           sample_writers[0], diagnostic_json_writers[0]);
     } else {
-      std::string output_file
-          = get_arg_val<string_argument>(parser, "output", "file");
-      auto base_sfx = get_basename_suffix(output_file);
-      if (base_sfx.second.empty())
-        base_sfx.second = ".csv";
-      auto ofs = std::make_unique<std::ofstream>(base_sfx.first + "_pathfinder"
-                                                 + base_sfx.second);
-      if (sig_figs > -1) {
-        ofs->precision(sig_figs);
-      }
-      stan::callbacks::unique_stream_writer<std::ofstream> pathfinder_writer(
-          std::move(ofs), "# ");
       write_stan(pathfinder_writer);
       write_model(pathfinder_writer, model.model_name());
       write_datetime(pathfinder_writer);
--- a/src/cmdstan/command.hpp
+++ b/src/cmdstan/command.hpp
@@ -195,10 +195,9 @@ int command(int argc, const char *argv[]) {
       diagnostic_csv_writers;
   std::vector<stan::callbacks::json_writer<std::ofstream>>
       diagnostic_json_writers;
-  stan::callbacks::unique_stream_writer<std::ofstream> pathfinder_writer;
 
   init_callbacks(parser, sample_writers, diagnostic_csv_writers,
-                 diagnostic_json_writers, pathfinder_writer);
+                 diagnostic_json_writers);
 
   // Setup initial parameter values - arg "init"
   // arg is either filename or init radius value
@@ -266,6 +265,12 @@ int command(int argc, const char *argv[]) {
           save_iterations, refresh, interrupt, logger, init_writer,
           sample_writers[0], diagnostic_json_writers[0]);
     } else {
+      std::vector<std::string> output_filenames =
+        make_filenames(get_arg_val<string_argument>(parser, "output", "file"), "", ".csv", 1, id);
+      auto ofs = std::make_unique<std::ofstream>(output_filenames[0]);
+      if (sig_figs > -1)
+        ofs->precision(sig_figs);
+      stan::callbacks::unique_stream_writer<std::ofstream> pathfinder_writer(std::move(ofs), "# ");
       write_stan(pathfinder_writer);
       write_model(pathfinder_writer, model.model_name());
       write_datetime(pathfinder_writer);
--- a/src/cmdstan/command.hpp
+++ b/src/cmdstan/command.hpp
@@ -267,10 +267,11 @@ int command(int argc, const char *argv[]) {
           save_iterations, refresh, interrupt, logger, init_writer,
           sample_writers[0], diagnostic_json_writers[0]);
     } else {
-      std::vector<std::string> pf_names = make_filenames(
-          get_arg_val<string_argument>(parser, "output", "file"), "", ".csv", 1,
-          id);
-      auto ofs = std::make_unique<std::ofstream>(pf_names[0]);
+      auto ofs
+          = std::make_unique<std::ofstream>(
+              get_basename_suffix(
+                  get_arg_val<string_argument>(parser, "output", "file")).first
+              + ".csv");
       if (sig_figs > -1)
         ofs->precision(sig_figs);
       stan::callbacks::unique_stream_writer<std::ofstream> pathfinder_writer(
--- a/src/cmdstan/command.hpp
+++ b/src/cmdstan/command.hpp
@@ -267,11 +267,11 @@ int command(int argc, const char *argv[]) {
           save_iterations, refresh, interrupt, logger, init_writer,
           sample_writers[0], diagnostic_json_writers[0]);
     } else {
-      auto ofs
-          = std::make_unique<std::ofstream>(
-              get_basename_suffix(
-                  get_arg_val<string_argument>(parser, "output", "file")).first
-              + ".csv");
+      auto ofs = std::make_unique<std::ofstream>(
+          get_basename_suffix(
+              get_arg_val<string_argument>(parser, "output", "file"))
+              .first
+          + ".csv");
       if (sig_figs > -1)
         ofs->precision(sig_figs);
       stan::callbacks::unique_stream_writer<std::ofstream> pathfinder_writer(
--- a/src/cmdstan/command.hpp
+++ b/src/cmdstan/command.hpp
@@ -208,9 +208,11 @@ int command(int argc, const char *argv[]) {
     init = "";
   } catch (const std::logic_error &e) {
   }
+
   std::vector<std::shared_ptr<stan::io::var_context>> init_contexts
       = get_vec_var_context(init, num_chains);
   std::vector<std::string> model_compile_info = model.model_compile_info();
+
   for (int i = 0; i < num_chains; ++i) {
     write_stan(sample_writers[i]);
     write_model(sample_writers[i], model.model_name());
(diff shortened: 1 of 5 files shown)


src/cmdstan/command.hpp
```diff
@@ -264,16 +265,12 @@ int command(int argc, const char *argv[]) {
...
-      auto base_sfx = get_basename_suffix(output_file);
-      if (base_sfx.second.empty())
-        base_sfx.second = ".csv";
-      auto ofs = std::make_unique<std::ofstream>(base_sfx.first + "_pathfinder"
-                                                 + base_sfx.second);
-      if (sig_figs > -1) {
+      std::vector<std::string> output_filenames = make_filenames(
+          get_arg_val<string_argument>(parser, "output", "file"), "", ".csv", 1,
+          id);
+      auto ofs = std::make_unique<std::ofstream>(output_filenames[0]);
```
WardBrian, line 271: Does this need the extra indirection with the helper function call and the vector return?

src/cmdstan/command_helper.hpp
```diff
@@ -767,22 +768,40 @@ void init_callbacks(
         &diag_csv_writers,
     std::vector<stan::callbacks::json_writer<std::ofstream>>
         &diag_json_writers) {
+  // bookkeeping
```
WardBrian, line 771:
```suggestion

```

src/cmdstan/command_helper.hpp
```diff
@@ -767,22 +768,40 @@ void init_callbacks(
...
-    sample_writers.emplace_back(std::move(ofs), "# ");
+  if (!is_pathfinder || save_single_paths || num_chains == 1) {
+    std::vector<std::string> output_filenames
+        = make_filenames(get_arg_val<string_argument>(parser, "output", "file"),
+                         "", ".csv", num_chains, id);
+    if (save_single_paths) {
+      output_filenames = make_filenames(
+          get_arg_val<string_argument>(parser, "output", "file"), "_path",
+          ".csv", num_chains, id);
+    }
```
WardBrian, line 792: This seems like it should be an if/else, rather than re-assigning to the vector

src/cmdstan/command_helper.hpp
```diff
@@ -767,22 +768,40 @@ void init_callbacks(
...
+    for (int i = 0; i < num_chains; ++i) {
+      auto ofs = std::make_unique<std::ofstream>(output_filenames[i]);
+      if (sig_figs > -1)
+        ofs->precision(sig_figs);
+      sample_writers.emplace_back(std::move(ofs), "# ");
+    }
+  } else if (is_pathfinder && !save_single_paths) {
+    for (int i = 0; i < num_chains; ++i) {
+      sample_writers.emplace_back(nullptr, "# ");
+    }
```
WardBrian, line 802: I'm happy to be overruled on this but I think the intention reads clearer if this special case is an if, and then everything else is just an 'else'

So you'd have (I think) `  if (is_pathfinder && !(save_single_paths && num_chains > 1) ) {` followed by this block, and then an `else` with all the above.

A comment explaining why we're branching would also be nice!

WardBrian (changes requested):

WardBrian:
Do we feel like this should interact with the diagnostic files at all? I know at the moment there are no multipath-pathfinder specific diagnostics in reality, so it might be odd to require the `save_single_paths` argument to get any diagnostics out

mitzimorris:
> it might be odd to require the save_single_paths argument to get any diagnostics out

but this isn't required - `save_single_paths` outputs CSV draws from the individual pathfinders.
the diagnostics files report on the Pathfinder trajectory via LBFGS.

WardBrian:
I understand the current behavior, I'm asking if we think it should have an impact

Imagine one day we actually use the argument for multi-pathfinder to have diagnostics (independent of the single path diagnostics). Then it would seem natural that this argument also controlled whether the single-path diagnostics were saved.

mitzimorris:
I see - so if you want the diagnostics file then you'd probably want the single-path pathfinders as well?
which means that we don't need "save_single_paths" - `diagnostic_file` arg controls everything?
and then you'd have both `.json` and `.csv` files which have the same basename?

very elegant!

WardBrian:
I was thinking of more or less the flip of that - what should this argument do if there are multiple kinds of diagnostics (like there are multiple kinds of output csvs at the moment)

mitzimorris:
add more arguments to pathfinder method that are fine-grained enough to handle this and forget about output arg `diagnostics_file`?

WardBrian:
I think I was unclear. 

This PR was motivated because the current “output” argument controls both the output of the PSIS draws and the individual paths. My point is just that one day, if we actually use the diagnostic file argument to the multi-path algorithm, we will have the same thing with the “diagnostic file” argument - some for the PSIS, some for single paths. Whether we care is another question, and the answer can be “no”

mitzimorris:
if I understand correctly proposal is this:

- `diagnostic_file` - is an output sub-argument which is reserved for multi-path Pathfinder.
- "save_single_paths" - for method `pathfinder` is a boolean arg which, if true, will create both the CSV file of draws and the json file of ELBO evaluations.

if I understand correctly, `save_single_paths` is useful mainly for exploring Pathfinder behavoir on challenging or ill-specified models.

WardBrian:
Currently, this PR implements:

| - | `output=foo` | `diagnostic_file=bar` |
|--------|--------|--------|
| **`save_single_paths=0`** | One file, `foo.csv`, containing PSIS draws | Several `bar_N.json` files, one for each path, containing single-pathfinder diagnostics |
| **`save_single_paths=1`** | The same `foo.csv` file, plus individual `foo_path_N.csv` files, one for each path | Several `bar_N.json` files, one for each path, containing single-pathfinder diagnostics |


My question was, should the matrix look instead like:

| - | `output=foo` | `diagnostic_file=bar` |
|--------|--------|--------|
| **`save_single_paths=0`** | One file, `foo.csv`, containing PSIS draws | One file, `bar.json`, containing the diagnostics for multi-path pathfinder (currently, this is completely unused). |
| **`save_single_paths=1`** | The same `foo.csv` file, plus individual `foo_path_N.csv` files, one for each path | The same `bar.json` file, plus individual `bar_path_N.json` files containing the diagnostics for each single-path |


As noted, there are currently no diagnostics for multi-path _besides_ each individual path's diagnostics, but if we ever think there would be then maybe the second table makes more sense. Of course, if someone is asking for diagnostics, maybe it's fine just to give them more than they asked for. It's a much less common use case than output, at any rate

mitzimorris:
two things:

- creating an empty multi-path pathfinder diagnostics file is confusing
-  having to specify diagnostic file basename could be skipped in favor of adding distinguishing tags for output file basename.

mitzimorris:
per discussion w/ @WardBrian and @SteveBronder:

- `output` sub-arg `diagnostic_file` will be used for outputs from multi-path Pathfinder.  currently, dummy json writer passed in to service method call.
-  `pathfinder` boolean arg `save_single_paths`, when true, will save both single-path sample as well as single-path diagnostics which report on the ELBO evaluations.   these files names will constructed from the `output_file` basename, plus `_path` + path id.

Question:  what should single path pathfinder output file be named if `num_paths= 1`?   what about `num_paths=1` and `save_single_paths=1`?   Choices for output filenames:

- a.  `basename_path_1.csv`  and `basename_path_1.json`
- b.  `basename.csv` and `basename.json`

I prefer option b.  opinions?  @bob-carpenter ?

WardBrian:
I think if num paths is 1 it should just be “output.csv”, and the diagnostic should probably come from the diagnostic file argument

mitzimorris:
how about this:

if `num_paths=1` and `save_single_paths=1` and arg `diagnostic_file` is unspecified, then single-path diagnostics are in file `output.json`.

if `num_paths=1` and `diagnostic_file="foo.json"`, then single-path diagnostics are in file `foo.json`.

mitzimorris:
this is ready for re-review.

mitzimorris:
CI failed on upstream/(downstream?) performance tests - one of the benchmarks gold files is missing?
@WardBrian @serban-nicusor-toptal 

```

Informational Message: The current Metropolis proposal is about to be rejected because of the following issue:
Exception: gp_exp_quad_cov: sigma is 0, but must be positive! (in '../stat_comp_benchmarks/benchmarks/gp_pois_regr/gp_pois_regr.stan', line 14, column 4 to line 15, column 59)
If this warning occurs sporadically, such as for highly constrained variable types like covariance matrices, then the sampler is fine,
but if this warning occurs often then your model may be either severely ill-conditioned or misspecified.

Iteration:  100 / 2000 [  5%]  (Warmup)
Iteration:  200 / 2000 [ 10%]  (Warmup)
Iteration:  300 / 2000 [ 15%]  (Warmup)
Iteration:  400 / 2000 [ 20%]  (Warmup)
Iteration:  500 / 2000 [ 25%]  (Warmup)
Iteration:  600 / 2000 [ 30%]  (Warmup)
Iteration:  700 / 2000 [ 35%]  (Warmup)
Iteration:  800 / 2000 [ 40%]  (Warmup)
Iteration:  900 / 2000 [ 45%]  (Warmup)
Iteration: 1000 / 2000 [ 50%]  (Warmup)
Iteration: 1001 / 2000 [ 50%]  (Sampling)
Iteration: 1100 / 2000 [ 55%]  (Sampling)
Iteration: 1200 / 2000 [ 60%]  (Sampling)
Iteration: 1300 / 2000 [ 65%]  (Sampling)
Iteration: 1400 / 2000 [ 70%]  (Sampling)
Iteration: 1500 / 2000 [ 75%]  (Sampling)
Iteration: 1600 / 2000 [ 80%]  (Sampling)
Iteration: 1700 / 2000 [ 85%]  (Sampling)
Iteration: 1800 / 2000 [ 90%]  (Sampling)
Iteration: 1900 / 2000 [ 95%]  (Sampling)
Iteration: 2000 / 2000 [100%]  (Sampling)

 Elapsed Time: 1.795 seconds (Warm-up)
               2.345 seconds (Sampling)
               4.14 seconds (Total)

Traceback (most recent call last):
  File "./runPerformanceTests.py", line 424, in <module>
    results = list(results)
  File "./runPerformanceTests.py", line 345, in process_test_wrapper
    time_, (fails, errors) = run(exe, data, overwrite, check_golds,
  File "./runPerformanceTests.py", line 275, in run
    summary = csv_summary(tmp)
  File "./runPerformanceTests.py", line 163, in csv_summary
    with open(csv_file, 'r', encoding = 'utf-8') as raw:
FileNotFoundError: [Errno 2] No such file or directory: 'golds/stat_comp_benchmarks_benchmarks_gp_pois_regr_gp_pois_regr.gold.tmp'
script returned exit code 1
```

WardBrian:
It’s because the performance tests request output files that end in .tmp:
https://github.com/stan-dev/performance-tests-cmdstan/blob/ffff830fbd6d5e09dbee67dad336e957e9e19b2a/runPerformanceTests.py#L268

But the latest changes here (to make_filenames specifically) always overwrite that to be .csv

src/cmdstan/command.hpp
```diff
@@ -264,16 +267,12 @@ int command(int argc, const char *argv[]) {
...
-      auto base_sfx = get_basename_suffix(output_file);
-      if (base_sfx.second.empty())
-        base_sfx.second = ".csv";
-      auto ofs = std::make_unique<std::ofstream>(base_sfx.first + "_pathfinder"
-                                                 + base_sfx.second);
-      if (sig_figs > -1) {
+      std::vector<std::string> pf_names = make_filenames(
+          get_arg_val<string_argument>(parser, "output", "file"), "", ".csv", 1,
+          id);
+      auto ofs = std::make_unique<std::ofstream>(pf_names[0]);
```
WardBrian, line 273: I think I still prefer the original code which just repeated the piece of logic needed rather than calling this function and creating a length-1 vector

src/cmdstan/command_helper.hpp
```diff
@@ -738,14 +738,13 @@ void check_file_config(argument_parser &parser) {
...
+                                        const std::string &tag,
                                         const std::string &type,
                                         unsigned int num_chains,
                                         unsigned int id) {
   std::vector<std::string> names(num_chains);
   auto base_sfx = get_basename_suffix(filename);
-  if (base_sfx.second.empty()) {
-    base_sfx.second = type;
-  }
+  base_sfx.second = type;
```
WardBrian, line 747: As noted, this is breaking tests (and would possibly break users code, if they were expecting the old behavior for filenames).

I think you could either add something like an `bool override_type = false` argument to keep the old behavior for all cases besides the pathfinder JSONs, or reconsider if reusing the output filename is the right abstraction

WardBrian (approved):

PR #1186, opened 2023-08-21
Add option to keep object file after compiling model

andrjohns:
#### Submisison Checklist

#### Summary:

This PR adds a makeflag `KEEP_OBJECT` to allow keeping the `.o` object file when compiling a model.

When exposing model methods (`log_prob`, etc) in `cmdstanr` we end up having to compile a given model twice - once via `cmdstan` to generate the executable and once via Rcpp to expose the `stan_model` typedef to generate a model pointer. If we can still have access to the object file, we can instead just link against it and skip the additional compilation.

#### Intended Effect:

Reduce unnecessary compilation when exposing model methods in `cmdstanr`

#### How to Verify:

Add `KEEP_OBJECT=true` to `make/local`

#### Side Effects:

N/A

#### Documentation:

N/A

#### Copyright and Licensing


mitzimorris:
will the makefile's "clean" methods get rid of these ".o" files?

WardBrian:
> will the makefile's "clean" methods get rid of these ".o" files?

The .o files can be left around anywhere on disk (not necessarily under the cmdstan folder) so they wouldn't be

WardBrian:
@andrjohns do you mind adding a generic `*.o` entry to our gitignore? We have one right now specifically for main.o, but if someone has this in their make/local it would become annoying for development etc.

Otherwise I think this can go in

andrjohns:
Obsoleted by #1187

PR #1203, opened 2023-09-18
Feature/1192 save metric json

mitzimorris:
#### Submisison Checklist

#### Summary:
Add option to adaptive sampler args to save metric in separate JSON file with elements "stepsize" and "inv_metric"

#### Intended Effect:
Easy way to get pre-specified metric.

#### How to Verify:
Unit tests will pass once https://github.com/stan-dev/stan/pull/3230 has been merged.

#### Side Effects:
N/A

#### Documentation:
Will add separately to docs repo.

#### Copyright and Licensing


src/cmdstan/command_helper.hpp
```diff
@@ -766,8 +766,9 @@ void init_callbacks(
         &sample_writers,
     std::vector<stan::callbacks::unique_stream_writer<std::ofstream>>
         &diag_csv_writers,
+    std::vector<stan::callbacks::json_writer<std::ofstream>> &diag_json_writers,
     std::vector<stan::callbacks::json_writer<std::ofstream>>
-        &diag_json_writers) {
+        &metric_json_writers) {
```
WardBrian, line 771: I think a fourth in-out parameter on this function is probably where we should draw the line and split it up or refactor it.

I think you can simplify this logic a lot by writing two helper functions:

```c++
template <typename T>
void null_initialize_callbacks(std::vector<T> &writers, size_t num_chains) {
  writers.reserve(num_chains);
  for (size_t i = 0; i < num_chains; ++i) {
    writers.emplace_back(nullptr);
  }
}

template <typename T, typename... Ts>
void file_initialize_writers(std::vector<T> &writers, unsigned int num_chains,
                             unsigned int id, std::string &filename,
                             std::string tag, std::string suffix, int sig_figs,
                             Ts &&...args) {
  writers.reserve(num_chains);
  auto filenames = make_filenames(filename, tag, suffix, num_chains, id);

  for (size_t i = 0; i < num_chains; ++i) {
    auto ofs = std::make_unique<std::ofstream>(filenames[i]);
    if (sig_figs > -1) {
      ofs->precision(sig_figs);
    }
    writers.emplace_back(std::move(ofs), std::forward<Ts>(args)...);
  }
}
```

The first will fill the callback lists with nullptrs, the second will create files. You could use them in the existing function like:

```
  // default - no diagnostics
  null_initialize_callbacks(diag_csv_writers, num_chains);
  null_initialize_callbacks(diag_json_writers, num_chains);
```

```
else {  // not pathfinder

    file_initialize_writers(sample_writers, num_chains, id, output_file, "",
                            ".csv", sig_figs, "# ");

```

But I think it might be better to break this function apart and inline parts of it using those helpers.

mitzimorris: changed per your suggestion.  many thanks!

mitzimorris:
this is ready for re-review

src/cmdstan/command.hpp
```diff
@@ -255,42 +306,30 @@ int command(int argc, const char *argv[]) {
...
-    bool save_iterations
-        = get_arg_val<bool_argument>(*pathfinder_arg, "save_single_paths");
-
-    if (num_paths == 1) {
-      if (!get_arg_val<string_argument>(parser, "output", "diagnostic_file")
-               .empty()) {
-        save_iterations = true;
-      }
+    if (num_chains == 1) {
+      save_single_paths = save_single_paths || !diagnostic_file.empty();
```
WardBrian, line 310: I don't understand this change

Are the diagnostic files ever nonempty if there's no saved single paths? I don't fully recall what we decided for the interface, might be worth commenting here.

src/cmdstan/command_helper.hpp
```diff
@@ -750,109 +750,28 @@ void check_file_config(argument_parser &parser) {
...
-  diag_csv_writers.reserve(num_chains);
-  diag_json_writers.reserve(num_chains);
-
-  // default - no diagnostics
-  for (int i = 0; i < num_chains; ++i) {
-    diag_csv_writers.emplace_back(nullptr, "# ");
-    diag_json_writers.emplace_back(
-        stan::callbacks::json_writer<std::ofstream>());
+template <typename T>
+void init_null_callbacks(std::vector<T> &writers, size_t num_chains) {
```
WardBrian, line 754: Glad these functions worked out, I think the result is very nice.

src/test/interface/command_test.cpp
WardBrian, line 1: Why did this test need to change?

src/test/interface/save_metric_json_test.cpp
```diff
@@ -0,0 +1,102 @@
...
+  ASSERT_FALSE(out.hasError);
+  ASSERT_TRUE(file_exists(convert_model_path(output_metric)));
+
+  std::fstream result_stream(convert_model_path(output_metric));
+  std::stringstream result_sstream;
+  result_sstream << result_stream.rdbuf();
+  result_stream.close();
+  std::string metric = result_sstream.str();
+  rapidjson::Document document;
+  ASSERT_FALSE(document.Parse<0>(metric.c_str()).HasParseError());
```
WardBrian, line 54: You can  `#include <test/unit/util.hpp>` and use `ASSERT_TRUE(stan::test::is_valid_JSON(...));` instead for these tests

src/test/interface/save_metric_json_test.cpp
```diff
@@ -0,0 +1,102 @@
...
+  EXPECT_EQ(count_matches("stepsize", metric), 1);
+  EXPECT_EQ(count_matches("inv_metric", metric), 1);
+  EXPECT_EQ(count_matches("[", metric), 7);  // diagonal metric
```
WardBrian, line 79:
```suggestion
  EXPECT_EQ(count_matches("[", metric), 7);  // dense metric
```

src/test/interface/save_metric_json_test.cpp
```diff
@@ -0,0 +1,102 @@
...
+  std::fstream result_stream(convert_model_path(output_metric));
+  std::stringstream result_sstream;
+  result_sstream << result_stream.rdbuf();
+  result_stream.close();
+  std::string metric = result_sstream.str();
+  rapidjson::Document document;
+  ASSERT_FALSE(document.Parse<0>(metric.c_str()).HasParseError());
+  EXPECT_EQ(count_matches("stepsize", metric), 1);
+  EXPECT_EQ(count_matches("inv_metric", metric), 1);
+  EXPECT_EQ(count_matches("[", metric), 1);  // diagonal metric
```
WardBrian, line 101:
```suggestion
  EXPECT_EQ(count_matches("[", metric), 1);  // unit metric is diagonal
```

WardBrian: In this case you also know what the metric would be, which you could test if you wanted

src/cmdstan/command.hpp
```diff
@@ -187,9 +193,61 @@ int command(int argc, const char *argv[]) {
...
+    }
+    init_null_callbacks(diagnostic_csv_writers, num_chains);
+  } else {
+    init_filestream_writers(sample_writers, num_chains, id, output_file, "",
+                            ".csv", sig_figs, "# ");
+    if (save_diagnostics) {
+      init_filestream_writers(diagnostic_csv_writers, num_chains, id,
+                              diagnostic_base, "", ".csv", sig_figs, "# ");
+    } else {
+      init_null_callbacks(diagnostic_csv_writers, num_chains);
```
WardBrian, line 239: [Comment] If we ever get uncomfortable with the amount of nesting in this config, we could write a `conditional_init_writers` function which takes in a bool and all the remaining arguments and does this common if-else pattern.

src/cmdstan/command_helper.hpp
WardBrian: Can we unify the naming between them (`callbacks` vs `writers` - I prefer `writers`) -- my bad!

WardBrian (commented):
This is looking really nice. A few comments but they're all minor now

src/cmdstan/command.hpp
mitzimorris: the logic is super-opaque - I don't quite remember either.   I think our reasoning was:

- save_single_paths always gives you the diagnostic file
-if somone specifies a diagnostic file and num_paths == 1,  give them a diagnostic file which has whatever basename they specify.
- if someone specifies num_paths==1, but doesn't want a diagnostic file, don't output one.

src/test/interface/command_test.cpp
mitzimorris: at one point I messed up the comment prefix - specified "`#`" instead of "`# `#.   the tests were failing, so changed to check for message, independent of prefix.    could be reverted - also makes sense as it is, no?

WardBrian: I think reverting makes sense to make sure we keep the prefix the same (since _some_ downstream thing probably depends on the comments having a space after `#`...)

src/cmdstan/command.hpp
mitzimorris: so far, just 2.   it's a toss up.

src/test/interface/save_metric_json_test.cpp
mitzimorris: yes, but string match won't work and parsing out the values from JSON is a PITA.

WardBrian (approved):

PR #1215, opened 2023-11-03
Use stringstream to format default values

### LINKED ISSUE #1214: Discrepancy between `help-all` print and defaults: Pathfinder, LBFGS and BFGS
#### Summary:
The detailed print of the argument tree contains a few nodes which exhibit a discrepancy between the default value (as printed) and the default logged in the output. This only affects `tol_obj`, `tol_grad` and `tol_param`.

#### Description:
The problem affects 2 methods: `optimize` and `pathfinder`. For `optimize`, both the LBFGS and BFGS algorithms are affected. Together, this forms 3 cases in which the same 3 `double` values are printed with insufficient precision by `help-all`.

#### Reproducible Steps:
Assuming that one has the Bernoulli example built:

```shell
# Produce the default values which are logged in the output
./bernoulli method=optimize algorithm=lbfgs data file=bernoulli.data.json output file=lbfgs.csv
./bernoulli method=optimize algorithm=bfgs data file=bernoulli.data.json output file=bfgs.csv
./bernoulli method=pathfinder data file=bernoulli.data.json output file=pathfinder.csv

for name in lbfgs bfgs pathfinder ; do
    echo ================ ${name} ================
    grep -E "tol_(obj|grad|param)" ${name}.csv
done

# Compare against defaults in detailed print -- `rg` is ripgrep (has multiline option)
./bernoulli help-all | rg -U "tol_(obj|grad|param)=<double>\n.*\n.*\n.*Defaults to"
```

#### Current Output:
For each case of the 3 cases, this produces the information in the table below.

| variable| default value printed by `help-all` | true default logged by call with no arguments supplied|
|---|---|---|
|`tol_obj`| `0.000000` |`9.9999999999999998e-13`|
|`tol_grad`| `0.000000` | `1e-08`|
|`tol_param`| `0.000000` | `1e-08`|

#### Expected Output:
`help-all` should print values with sufficient precision.

#### Current Version:
v2.33.1

WardBrian:
#### Submisison Checklist

#### Summary:

We use `std::to_string` to format default values (since #1172) but this does not print floating point numbers in scientific notation when required, see the [example on cppreference](https://en.cppreference.com/w/cpp/string/basic_string/to_string#Example) comparing `operator<<` and `std::to_string`

#### Intended Effect:

Closes #1214 

#### How to Verify:

`./bernoulli help-all | grep -E "tol_(obj|grad|param)=<double>" -A 3`

#### Copyright and Licensing

### DIFF
--- a/src/cmdstan/arguments/arg_single_real_bounded.hpp
+++ b/src/cmdstan/arguments/arg_single_real_bounded.hpp
@@ -3,6 +3,7 @@
 
 #include <cmdstan/arguments/singleton_argument.hpp>
 #include <string>
+#include <sstream>
 
 /** Generic bounded real value argument */
 
@@ -21,7 +22,10 @@ class arg_single_real_bounded : public real_argument {
     _validity
         = std::to_string(lb).append(" <= ").append(name).append(" <= ").append(
             std::to_string(ub));
-    _default = std::to_string(def);
+
+    std::stringstream def_ss;
+    def_ss << def;
+    _default = def_ss.str();
     _default_value = def;
     _value = _default_value;
     _lb = lb;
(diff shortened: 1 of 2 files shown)


mitzimorris (approved):
lgtm!

PR #1222, opened 2023-11-09
Replace usage of killOldBuilds with disableConcurrentBuilds

serban-nicusor-toptal:
#### Submisison Checklist

#### Summary:

Due to some Jenkins changes, we have to migrate away from the old way of killing concurrent builds.

#### Copyright and Licensing

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -46,7 +46,10 @@ String branchName() { isPR() ? env.CHANGE_BRANCH :env.BRANCH_NAME }
 
 pipeline {
     agent none
-    options { skipDefaultCheckout() }
+    options {
+        skipDefaultCheckout()
+        disableConcurrentBuilds(abortPrevious: true)
+    }
     parameters {
         string(defaultValue: '', name: 'stan_pr',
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -67,14 +70,6 @@ pipeline {
         GIT_COMMITTER_EMAIL = 'mc.stanislaw@gmail.com'
     }
     stages {
-        stage('Kill previous builds') {
-            when {
-                not { branch 'develop' }
-                not { branch 'master' }
-                not { branch 'downstream_tests' }
```
WardBrian, line 74: I think this actually will cause problems, since we use `downstream_tests` for multiple other pipelines, and this will kill all-but-the-latest of them.

@dylex - you mentioned there was a way to make this more configurable?

WardBrian (approved):

Jenkinsfile
dylex: Looks like you figured it out, but here's a similar example from another project:
```
def keep = !env.BRANCH_NAME.startsWith("PR-")
properties([pipelineTriggers(keep ? [...] : [])])
```

Issue #1213, opened 2023-11-02
pathfinder: output files don't respect multiple periods, clobber outputs


WardBrian:
#### Summary:

Reported by @andrewjradcliffe

Passing an output file argument with multiple `.`s in it and requesting `save_single_paths` leads to output files with the same name, so they all end up clobbering the output files on disk.

#### Reproducible Steps:

```
./bernoulli pathfinder save_single_paths=1 output file=foo.bar.baz data file=bernoulli.data.json
```

#### Current Output:

This produces files `foo.bar.csv  foo_path_1.bar  foo_path_2.bar  foo_path_3.bar  foo_path_4.bar`

The files `foo_path_1.bar`... all have a incoherent mix of json/csv in them.

#### Additional Information:

I believe this is because the `make_filenames` function doesn't replace an extension if it is already provided. This works if there is 1 `.` in the name since we artificially strip it out for pathfinder, but if there is more than one we get this problem. 

#### Current Version:
v2.33.1

mitzimorris:
the logic is all fubar.   for the downstream regression tests, there's one set that uses suffix ".tmp" instead of ".csv".

this logic would be greatly simplified if we specified that output CSV files are always suffixed ".csv" and changed the regression testset names.

WardBrian:
> this logic would be greatly simplified if we specified that output CSV files are always suffixed ".csv" and changed the regression testset names.

That seems fine to me? Maybe we should print a warning if they were specified with another extension and we ignore it. Pathfinder is technically the only interface where we need to be able to ignore it, since the csvs and jsons differ only by extension, but threading a flag through the helper code seems like overkill

Issue #1226, opened 2023-12-07
More output file name quirks


andrewjradcliffe:
#### Summary:
As noted in [this issue](https://github.com/stan-dev/cmdstan/issues/1213), the treatment of paths as strings in CmdStan can lead to strange behavior.

#### Description:
Treating a string as a path, and subsequently [splitting on last dot](https://github.com/stan-dev/cmdstan/blob/35d1eab68e1823416e0f249e6ddf9a8d46b9c229/src/cmdstan/command_helper.hpp#L115) produces a family of problems.

#### Reproducible Steps:
I am certain that more examples can be produced, but here are a few interesting cases.

Assuming that you have the bernoulli example built and your current directory is `examples/bernoulli/bernoulli`:

##### Case 1
```bash
rm -rf foo* && mkdir -p foo.bar

# produces a file "baz" in directory "foo.bar"
./bernoulli data file=bernoulli.data.json sample num_chains=1 output file=foo.bar/baz
ls -a foo.bar

# produces no files
./bernoulli data file=bernoulli.data.json sample num_chains=3 output file=foo.bar/baz
ls -a foo.bar
```

##### Case 2
```bash
rm -rf foo* && mkdir -p foo/.bar

# produces a file "baz" in directory "foo/.bar"
./bernoulli data file=bernoulli.data.json sample num_chains=1 output file=foo/.bar/baz
ls -a foo/.bar

# produces no files
./bernoulli data file=bernoulli.data.json sample num_chains=3 output file=foo/.bar/baz
ls -a foo/.bar
```

##### Debug of Case 2
CmdStan is splitting on the `'.'` and producing output file names of the form `foo/_${id}.bar/baz` and since `_${id}.bar` is a non-existent directory, no files are created. If we create said directories ahead of time, the writer succeeds.

```bash
rm -rf foo* && mkdir -p foo/_1.bar foo/_2.bar

# Creates "foo/_1.bar/baz" and "foo/_2.bar/baz" but not "foo/_3.bar/baz"
./bernoulli data file=bernoulli.data.json sample num_chains=3 output file=foo/.bar/baz
ls -a foo/.bar
```

##### Case 3
```bash
rm -rf foo* && mkdir -p foo/bar

# produces no file(s); ".." is not a writable file
./bernoulli data file=bernoulli.data.json sample num_chains=1 output file='foo/bar/..'
ls -a foo
ls -a foo/bar

[...]

PR #1228, opened 2024-01-02
Issue/1226 filename parsing - some path logic.

mitzimorris:
#### Submisison Checklist

#### Summary:

Make parsing of filename suffixes more robust - adds logic to handle filepath separators plus some edge cases.

#### Intended Effect:

Correct weirdness reported in #1226 

#### How to Verify:

Unit tests

#### Side Effects:

N/A

#### Documentation:

N/A

#### Copyright and Licensing


mitzimorris:
@WardBrian or @dylex - could you please check on why this failed?

WardBrian:
I think it is because you checked in a change to the `stan` submodule which seems to point to a non-existent commit

src/cmdstan/command_helper.hpp
```diff
@@ -716,6 +780,22 @@ unsigned int get_num_chains(argument_parser &parser) {
 void check_file_config(argument_parser &parser) {
   std::string sample_file
       = get_arg_val<string_argument>(parser, "output", "file");
+  if (sample_file[sample_file.size() - 1] == get_path_separator()
+      || boost::algorithm::ends_with(sample_file, "..")) {
```
WardBrian, line 784: I think this could be its own helper `bool is_valid_filename` or similar

src/test/interface/command_helper_test.cpp
```diff
@@ -0,0 +1,155 @@
...
+  valid_arguments.push_back(new cmdstan::arg_output());
+  argument_parser parser(valid_arguments);
+  stan::callbacks::stream_writer info(std::cout);
+  stan::callbacks::stream_writer err(std::cerr);
+  int err_code = parser.parse_args(argv.size(), argv_prime, info, err);
+
+  EXPECT_NO_THROW(check_file_config(parser));
+}
+
+TEST(CommandHelper, check_filename_config_fubar) {
```
WardBrian, line 88:
```suggestion
TEST(CommandHelper, check_filename_config_bad) {
```

Perhaps nitpicky, but we use `bad`, `bad2`, etc other places so I think it would be more searchable if these had their names changed

src/cmdstan/command_helper.hpp
```diff
@@ -103,6 +102,64 @@ inline constexpr auto get_arg_val(List &&arg_list, Args &&... args) {
...
+ */
+char get_path_separator() {
+#if defined(WIN32) || defined(_WIN32) \
+    || defined(__WIN32) && !defined(__CYGWIN__)
+  static char path_separator = '\\';
+#else
+  static char path_separator = '/';
+#endif
+  return path_separator;
+}
```
WardBrian, line 119: This could be a compile-time constant rather than a function

stan
WardBrian, line 1: I believe this is what is breaking CI. Does this branch rely on any changes to the stan submodule?

WardBrian (changes requested):
Thanks for tackling. How difficult do you think it would be to also fix https://github.com/stan-dev/cmdstan/issues/1213?

stan
mitzimorris: no changes to the submodule.

WardBrian: Ok, I believe restoring the submodule and pushing should unblock CI

mitzimorris:
> How difficult do you think it would be to also fix https://github.com/stan-dev/cmdstan/issues/1213?

not difficult at all - will add to this PR.

stan
mitzimorris: restored?

```
git submodule update --init --recursive --checkout
Submodule path 'stan/lib/stan_math': checked out 'ffeedb5b5d882e5be0c09999089060f16fa8d8e9'
```

mitzimorris: that didn't work.  wtf?

mitzimorris:
checks are failing because performance tests specify output files with suffix ".tmp". 
submitted PR to cmdstan perf tests repo:  https://github.com/stan-dev/performance-tests-cmdstan/pull/62

@WardBrian - will this work?

src/cmdstan/command_helper.hpp
```diff
@@ -186,11 +259,17 @@ std::vector<std::string> make_filenames(const std::string &filename,
...
                                         unsigned int num_chains,
                                         unsigned int id) {
-  std::vector<std::string> names(num_chains);
-  auto base_sfx = get_basename_suffix(filename);
-  if (base_sfx.second.empty()) {
+  std::pair<std::string, std::string> base_sfx;
+  std::string fname = filename;
+  std::transform(fname.begin(), fname.end(), fname.begin(), ::tolower);
+  if (boost::algorithm::ends_with(fname, ".csv")
+      || boost::algorithm::ends_with(fname, ".json")) {
```
WardBrian, line 266: I'm not sure I understand why these are hard coded?

This does seem to fix the example provided in #1213, at least such that output files are properly generated (the `.baz` portion vanishes into thin air, but I suppose this is fine?), there is definitely still some oddities around though. For example,

```
./bernoulli pathfinder save_single_paths=1 output file=foo.bar.baz. data file=bernoulli.data.json
``` 
(note the period at the end of the output file) generates the same files as 
```
./bernoulli pathfinder save_single_paths=1 output file=foo.bar.baz data file=bernoulli.data.json
```

I think some more tests of specifically this function would be helpful!

bob-carpenter:
Could someone please point me to or write doc for how this is supposed to behave?  This PR just says "Make parsing of filename suffixes more robust" and the issue #1226 just lists a bunch of quirky behavior without a spec for what should happen in general.

The problem starts with the name of the argument, `output file=foo/.bar/baz`, because we use this argument to produce an output *directory* and output *base file name prefix*.  If we just took those two things as arguments, we wouldn't need any string processing at all and it'd be clear what's going on to the user.

To allow the current behavior, I would suggest (a) splitting directory off, and (b) operating on the file to remove the final period and anything following it.

The code needs to check that the string-based path is reasonable.  That is, we can't just have `foo/.` as input as that leaves no base file.  We can't have * in the path, etc.  Something needs to validate these are truly paths.

If it didn't break backward compatibility, I would say just slap another `.csv` onto the end of `foo/bar.csv` and be done with it.

mitzimorris:
made all suggested changes, added fix for #1213  - ready for re-review.

mitzimorris:
> Could someone please point me to or write doc for how this is supposed to behave? This PR just says "Make parsing of filename suffixes more robust" and the issue https://github.com/stan-dev/cmdstan/issues/1226 just lists a bunch of quirky behavior without a spec for what should happen in general.

- CmdStan output filename behavoirs
       -  if running a single chain,  the CSV output filename is whatever the `file` argument specifies;  however, if the filename doesn't have suffix `csv` this will be added.
       - if running multiple chains, then the chain id is appended to the end of the filename, e.g. `output_1.csv`
       - for Pathfinder, there's a "save-single-paths" option, this creates a set of per-chain CSV samples which include both tag "_path" as well as chain id - e.g., PSIS sample:  `output.csv`,  plus per-chain `output_path_1.csv`, etc.
       - for Pathfinder, the diagnostic file is JSON format, for the sampler the diagnostic file is CSV format.
     
The CmdStan `output` argument allows 2 sub-arguments:
- `file` - this is the CSV file of draws from the sampler
-  `diagnostic file` - added specifically for the sampler diagnostics and gradients in CSV format

This design (sic) is problematic because:

- multi-chain single-process runs need to use the specified output file names as templates and hack in chain id information
- it was never stipulated that output filenames should end in ".csv"
- Pathfinder can output both the PSIS sample and the inidividual pathfinder samples.
- the diagnostic file from Pathfinder is JSON, not CSV.

W/R/T creating good output filenames:

- this PR fixes the problem that directory names which contain '.' were being parsed as suffixes.
- this PR tries to address problems of specified filenames that are actually just a filepath (per reported quirkiness)
- output file suffixes should be either ".csv" or ".json", according to the output file type.


>The problem starts with the name of the argument, output file=foo/.bar/baz, because we use this argument to produce an output directory and output base file name prefix. If we just took those two things as arguments, we wouldn't need any string processing at all and it'd be clear what's going on to the user.

absolutely.   CmdStanPy only has argument `output_dir` and it handles the logic of naming all the different kinds of output files with unique and consistent names.

>To allow the current behavior, I would suggest (a) splitting directory off, and (b) operating on the file to remove the final period and anything following it.

That's pretty much the way that things are re-imlemented here.  The problem is that C++14 doesn't have good filepath parsing routines and the Boost `filesystem` library is not header-only.  However, C++17 will have a  `filesystem` library, so eventually this will be easier to support.


>The code needs to check that the string-based path is reasonable. That is, we can't just have foo/. as input as that leaves no base file. We can't have * in the path, etc. Something needs to validate these are truly paths.

Agreed.

>If it didn't break backward compatibility, I would say just slap another .csv onto the end of foo/bar.csv and be done with it.

Backwards compatibility is how we got here.  We just fixed an annoying feature of the downstream performance tests which created output files that ended in ".tmp" not ".csv".

src/cmdstan/command_helper.hpp
mitzimorris: more unit tests have been added.
more discussion below.

mitzimorris: I don't think that ".baz" will disappear into the air - see line 95 or `command_helper_test.cpp`

mitzimorris:
current behavoir:

```
/examples/bernoulli/bernoulli sample output file=foo.csv. data file=examples/bernoulli/bernoulli.data.json 
```
results in output file `foo.csv..csv`

and 

```
./examples/bernoulli/bernoulli sample output file=foo.bar. data file=examples/bernoulli/bernoulli.data.json
```
results in output file `foo.bar..csv`

WardBrian:
> If we just took those two things as arguments, we wouldn't need any string processing at all and it'd be clear what's going on to the user.

I think that even this would have some subtle issues, like if the 'base file prefix' also contained a path separator. 

I think the fundamental flaw is we should never have set it up like this in the first place. IMO, when multiple files are required, we should have required the user to give us a comma separated list of the correct length and removed any guesswork about "if I specify X as the argument, the actual files created on disk will have the name..."


> To allow the current behavior, I would suggest (a) splitting directory off, and (b) operating on the file to remove the final period and anything following it.

I think this is reasonable. We are also within our rights to say "besides for CSV outputs, we will force the correct extension". So the bug reported in #1213 would yield `foo.bar_path_1.baz` as a CSV file, but `foo.bar_path_1.json` as the diagnostic file. That would maintain backward compatibility (by allowing csv outputs to still have arbitrary extensions) but be a bit more sane going forward with new formats

src/cmdstan/command_helper.hpp
WardBrian: After pulling, and running `./bernoulli pathfinder save_single_paths=1 output file=foo.bar.baz data file=bernoulli.data.json`:

```
$ ls
bernoulli            bernoulli.data.R  foo.bar.csv         foo.bar_path_1.json  foo.bar_path_2.json  foo.bar_path_3.json  foo.bar_path_4.json
bernoulli.data.json  bernoulli.stan    foo.bar_path_1.csv  foo.bar_path_2.csv   foo.bar_path_3.csv   foo.bar_path_4.csv
```

no `baz` to be found! This is concerning, because I agree the test suggests otherwise!

WardBrian: This is caused by the `output_base` variable in `command.hpp`. Integration test may be needed, sadly

src/cmdstan/command.hpp
```diff
@@ -197,34 +197,33 @@ int command(int argc, const char *argv[]) {
...
         && (get_arg_val<bool_argument>(parser, "method", "pathfinder",
                                        "save_single_paths"));
-  bool save_diagnostics = diagnostic_base != output_base;
 
   if (user_method->arg("pathfinder")) {
+    if (save_single_paths && diagnostic_file.empty()) {
+      diagnostic_file = output_file;
+    }
+    std::cout << "save_single_paths ? " << save_single_paths
+              << " diagnostic file? " << diagnostic_file << std::endl;
```
WardBrian, line 210:
```suggestion
```

src/cmdstan/command.hpp
```diff
@@ -197,34 +197,33 @@ int command(int argc, const char *argv[]) {
...
         init_filestream_writers(diagnostic_json_writers, num_chains, id,
-                                diagnostic_base, "", ".json", sig_figs);
+                                diagnostic_file, "", ".json", sig_figs);
       } else {
         init_null_writers(diagnostic_json_writers, num_chains);
       }
     } else {
-      if (save_single_paths || save_diagnostics) {
-        init_filestream_writers(sample_writers, num_chains, id, output_base,
+      if (save_single_paths) {
```
WardBrian, line 222: To get exactly the same behavior as before with these arguments, you need to still have an or condition here

```suggestion
      if (save_single_paths || !diagnostic_file.empty()) {
```

src/cmdstan/command.hpp
```diff
@@ -338,7 +336,7 @@ int command(int argc, const char *argv[]) {
           save_single_paths, refresh, interrupt, logger, init_writer,
           sample_writers[0], diagnostic_json_writers[0]);
     } else {
-      auto ofs = std::make_unique<std::ofstream>(output_base + ".csv");
+      auto ofs = std::make_unique<std::ofstream>(output_file);
```
WardBrian, line 339: using make_filenames and taking the first element should make the behavior more consistent with other methods

src/test/interface/pathfinder_test.cpp
```diff
@@ -18,7 +18,7 @@ class CmdStan : public testing::Test {
     eight_schools_data
         = {"src", "test", "test-models", "eight_schools.data.json"};
     empty_model = {"src", "test", "test-models", "empty"};
-    arg_output = {"test", "output"};
+    arg_output = {"test", "output.csv"};
```
WardBrian, line 21: Revert after making above change

WardBrian (changes requested):
A few small things, should be last review. Thanks!

WardBrian (approved):
Thanks again for tackling!

PR #1231, opened 2024-01-04
Add version header for CmdStan

andrjohns:
#### Submisison Checklist

#### Summary:

This PR adds a `version.hpp` header following the conventions of the [Stan](https://github.com/stan-dev/stan/blob/develop/src/stan/version.hpp) and [Math](https://github.com/stan-dev/math/blob/develop/stan/math/version.hpp) libraries

#### Intended Effect:

The CmdStan version is currently only recorded in the [Makefile](https://github.com/stan-dev/cmdstan/blob/2a9a6be331519d0f66c5e0a600070f372774c8fe/makefile#L152), which means that any downstream users need to also include the makefiles in order to identify the version of the CmdStan headers & sources.

This allows downstream users to identify the version of the CmdStan headers/sources without needing to include & parse the makefile, improves consistency with the Stan & Math libraries

#### Copyright and Licensing

### DIFF
--- a/src/cmdstan/version.hpp
+++ b/src/cmdstan/version.hpp
@@ -1,0 +1,31 @@
+#ifndef CMDSTAN_VERSION_HPP
+#define CMDSTAN_VERSION_HPP
+
+#include <string>
+
+#ifndef STAN_STRING_EXPAND
+#define STAN_STRING_EXPAND(s) #s
+#endif
+
+#ifndef STAN_STRING
+#define STAN_STRING(s) STAN_STRING_EXPAND(s)
+#endif
+
+#define CMDSTAN_MAJOR 2
+#define CMDSTAN_MINOR 33
+#define CMDSTAN_PATCH 1
+
+namespace cmdstan {
+
+/** Major version number for CmdStan package. */
+const std::string MAJOR_VERSION = STAN_STRING(CMDSTAN_MAJOR);
+
+/** Minor version number for CmdStan package. */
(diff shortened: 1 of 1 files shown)


WardBrian (approved):
Seems useful. Thoughts on having this be the version we put in the CSV files? Code is here:
https://github.com/stan-dev/cmdstan/blob/develop/src/cmdstan/write_stan.hpp

WardBrian:
Also tagging @serban-nicusor-toptal so he would know that there is a new file that needs updating each release!

rok-cesnovar:
> Seems useful. Thoughts on having this be the version we put in the CSV files? 

I don't see a case where the cmdstan and stan versions would diverge, but it depends on what we want to write in the file, the Stan version of the interface version. I think the Stan version?

WardBrian:
> I don't see a case where the cmdstan and stan versions would diverge

They've diverged right now (CmdStan is on 2.33.1, Stan on 2.33.0) because the latest bugfix release didn't require any changes to Stan or Math

rok-cesnovar:
Oh right. I don't really have a preference honestly, was just being nitpicky :)

WardBrian:
I think given a cmdstan version, there is always a unique version of Stan shipped with that (we never update Stan without updating CmdStan) but the inverse need not be true (e.g., right now), so it seems reasonable to include the cmdstan version in the csv. It's not a big deal either way, I think

PR #1245, opened 2024-01-26
Properly allow multiple downstream jenkins runs

WardBrian:
See https://github.com/stan-dev/stan/pull/3262/files


serban-nicusor-toptal (approved):
Nice catch, thanks!

PR #1256, opened 2024-03-13
Use parallel standalone generated quantities

SteveBronder:
#### Submisison Checklist

#### Summary:

Note the line change # here is just from adding the test files. Theres about 100 code line changes in total

This swaps out the old `generated_quantities` function call for the new parallel version.

This is pretty straightforward besides the changes to `get_num_chains`. For generated quantities, we don't have the user pass the number of chains, so we need to do some deduction to figure out how many chains the program will be running. We use similar logic to how we figure out the number of inits. If the user supplies just a `my_file_name.csv` then we assume it's just 1 chain. But if a user supplies `my_file_name_{id}.csv` we search for `my_file_name_{{id} + 1, {id} _ 2,..., {id} + N}` sequentially until we cannot find anymore files. The total number of files is the total number of chains.

#### Intended Effect:

Allow users to call generated quantities in parallel.

#### How to Verify:

I only wrote one test in `generated_quantities_test.cpp` that checks that the program does not error out. Is that sufficient?

#### Documentation:

Docs added for new file `file::exists()`

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company): Simon's Foundation


src/cmdstan/command_helper.hpp
```diff
@@ -609,14 +609,40 @@ void services_log_prob_grad(const stan::model::model_base &model, bool jacobian,
...
  * @return int num chains or paths
  */
-unsigned int get_num_chains(argument_parser &parser) {
+inline unsigned int get_num_chains(argument_parser &parser, unsigned int id) {
   auto user_method = parser.arg("method");
   if (user_method->arg("pathfinder"))
     return get_arg_val<int_argument>(parser, "method", "pathfinder",
                                      "num_paths");
 
+  auto gq_arg = user_method->arg("generate_quantities");
```
WardBrian, line 621: Not crazy about this detection method - thoughts on gating this behavior behind a new `num_chains` argument similar to how multithreaded sampling works?

WardBrian (changes requested):

src/cmdstan/command_helper.hpp
SteveBronder: Yeah I like that idea a lot more

src/cmdstan/command.hpp
```diff
@@ -361,20 +360,33 @@ int command(int argc, const char *argv[]) {
   } else if (user_method->arg("generate_quantities")) {
     // ---- generate_quantities start ---- //
     auto gq_arg = parser.arg("method")->arg("generate_quantities");
+    // TODO: MOVE THIS UP
```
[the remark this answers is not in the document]
SteveBronder: Oh this is old and needs deleted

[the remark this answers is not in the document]
SteveBronder: Oh yes need to throw an error here for nonexisting files and mismatches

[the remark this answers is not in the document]
SteveBronder: Now needs to be deleted

src/test/interface/generated_quantities_test.cpp
```diff
@@ -60,6 +63,18 @@ TEST_F(CmdStan, generate_quantities_good) {
   ASSERT_FALSE(out.hasError);
 }
 
+TEST_F(CmdStan, generate_quantities_good_multi) {
+  std::stringstream ss;
+  ss << convert_model_path(bern_gq_model)
+     << " data file=" << convert_model_path(bern_data)
+     << " output file=" << convert_model_path(dev_null_path)
+     << " method=generate_quantities fitted_params="
+     << convert_model_path(bern_fitted_params_multi) << " num_chains=4";
```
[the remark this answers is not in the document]
SteveBronder: > I assume the system will work fine if num_chains is less than the number of files, nothing we can do about that, but what if it is more?

cmdstan will still generate `file_name_{1..N}`, but then when those are passed to parse_read_csv` an error will be thrown by `safe_open` when the file does not exist. That doesn't feel like a thing we need to test here

> As a subset of that, what if I have fitted_params_1.csv and fitted_params_3.csv but not _2, with num_chains=3

Same thing as above. cmdstan will tell you the "Unable to open `fitted_params_2.csv`

WardBrian:
@SteveBronder your last commit here deleted a variable that seems to still be used elsewhere. It would be nice to get this into 2.35 if you can take a quick look

WardBrian (changes requested):
1. The argument parsing of the fitted_params filename expects the `_#` suffix, which is a bit odd. I think the same argument name for `sample output file` and `generate_quantities fitted_params` should work
2. There is no check that the input filenames and the output filenames don't clobber, so you can overwrite your inputs and get an error about failing to read params

WardBrian (commented):
Thanks. The behavior now seems correct, just some comments on code redundancy etc

src/cmdstan/file.hpp
```diff
@@ -105,6 +105,51 @@ std::pair<std::string, std::string> get_basename_suffix(
...
+ * Check if two file paths are the same file.
+ * @note This function only handles very basic access patterns.
+ *  It will miss cases such as
+ *  path1: ./a/b/../b/file.txt
+ *  path1: ./a/b/file.txt
+ * @param path1 first file path
+ * @param path2 second file path
+ * @return true if paths are the same file, false otherwise
+ */
+bool check_same_file(const std::string &path1, const std::string &path2) {
```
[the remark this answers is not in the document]
SteveBronder: Added some tests. I also changed the name to `check_approx_same_file` because imo it makes it more clear that we are not doing anything fancy to verify the path is 100% correct. (another reason to move up to c++17 imo)

src/cmdstan/command.hpp
```diff
@@ -363,18 +362,33 @@ int command(int argc, const char *argv[]) {
...
     if (fname.empty()) {
-      msg << "Missing fitted_params argument, cannot run generate_quantities "
-             "without fitted sample.";
-      throw std::invalid_argument(msg.str());
+      throw std::invalid_argument(
+          "Missing fitted_params argument, cannot run generate_quantities "
+          "without fitted sample.");
+    } else if (file::check_same_file(fname, output_file)) {
+      throw std::invalid_argument(
+          "Output file cannot be the same as the input file.");
```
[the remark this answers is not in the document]
SteveBronder: Moved the test into here

src/cmdstan/file.hpp
```diff
@@ -105,6 +105,60 @@ std::pair<std::string, std::string> get_basename_suffix(
...
+ *  path1: ./a/b/../b/file.txt
+ *  path1: ./a/b/file.txt
+ * @param path1 first file path
+ * @param path2 second file path
+ * @return true if paths are the same file, false otherwise
+ */
+bool check_approx_same_file(const std::string &path1,
+                            const std::string &path2) {
+  std::cout << "\tpath1: " << path1 << std::endl;
+  std::cout << "\tpath2: " << path2 << std::endl;
```
WardBrian, line 121: Remove debugging

src/cmdstan/file.hpp
```diff
@@ -105,6 +105,60 @@ std::pair<std::string, std::string> get_basename_suffix(
...
+  const auto path1_size = path1.size();
+  const auto path2_size = path2.size();
+
+  std::cout << "\tpath1_size: " << path1_size << std::endl;
+  std::cout << "\tpath2_size: " << path2_size << std::endl;
```
WardBrian, line 126: Remove debugging

src/test/interface/file_test.cpp
```diff
@@ -213,3 +213,16 @@ TEST(CommandHelper, check_filename_config_bad3) {
 
   EXPECT_THROW(check_file_config(parser), std::invalid_argument);
 }
+
+TEST(CommandHelper, check_same_file_test) {
```
WardBrian, line 217: Thanks for the test. Adding at least one `EXPECT_TRUE` seems like a good idea to show the cases it does actually handle. This test would pass with a function that always returned false 😉

WardBrian (commented):
Two minor things and then should be good to merge.

runCmdStanTests.py
```diff
@@ -130,7 +131,7 @@ def runTest(name, mpi=False, j=1):
     executable = mungeName(name).replace('/',os.sep)
     xml = mungeName(name).replace(WIN_SFX, '')
     command = '%s --gtest_output="xml:%s.xml"' % (executable, xml)
-    if mpi:
+    if mpi is True:
```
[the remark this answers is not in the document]
SteveBronder: Had a trickle down bug that caused this. Fixed now

src/cmdstan/file.hpp
SteveBronder: Oops! My bad

SteveBronder:
@WardBrian I think I fix up the issues

runCmdStanTests.py
```diff
@@ -144,12 +145,18 @@ def main():
...
 
+    stan_mpi = False
     try:
         with open('make/local') as f:
-            stan_mpi =  'STAN_MPI' in f.read()
+            for (_, line) in enumerate(f):
+                if line.strip().startswith("#"):
+                    continue
+                else:
+                    stan_mpi = line.find('STAN_MPI=true') != -1
```
WardBrian, line 155: [not sure if it's important] This won't accurately detect things like `STAN_MPI=1`, which also enable MPI

WardBrian (approved):

runCmdStanTests.py
SteveBronder: It wasn't detecting that before. I think if we want to do that I'd rather do it in a different PR

WardBrian: It was searching for STAN_MPI, so it would have. you've just traded false positives for false negatives here

SteveBronder: Oh I thought you meant an environment variable the user set. I can open another PR to just make this search for `STAN_MPI`

PR #1171, opened 2023-07-06
Add CXX version to precompiled header name

WardBrian:
This idea was discussed with @rok-cesnovar here: https://github.com/roualdes/bridgestan/issues/151#issuecomment-1624241349

#### Submisison Checklist

- [x] Run tests: `./runCmdStanTests.py src/test`
- [x] Declare copyright holder and open-source license: see below

#### Summary:

This adds `CXX_MAJOR` and `CXX_MINOR` to the name of the precompiled header file when it is enabled. The goal is to prevent the annoying errors from clang updates etc.

#### Intended Effect:

Prevent clang updates from breaking PCH builds, see [cmdstanpy issues](https://github.com/stan-dev/cmdstanpy/issues?q=is%3Aissue+pch)

#### Side Effects:

This could lead to several PCHs being made over time if the user doesn't run a `make clean-all` every now and then. This could cause the directory size to increase over time for an installation.

#### Documentation:

None

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company):

Simons Foundation

By submitting this pull request, the copyright holder is agreeing to license the submitted work under the following licenses:
- Code: BSD 3-clause (https://opensource.org/licenses/BSD-3-Clause)
- Documentation: CC-BY 4.0 (https://creativecommons.org/licenses/by/4.0/)

PR #1268, opened 2024-05-09
Fix GCC's ability to find the precompiled header

WardBrian:
#### Submisison Checklist

#### Summary:

In #1171 we added the version of the C++ compiler to the precompiled header file name. This works for clang, since we have to explicitly pass `-include-pch path/to/header`, but GCC does an implicit lookup. Since the filename didn't match, it failed. 

This paragraph in the GCC manual saves us, though:

> If you need to precompile the same header file for different languages, targets, or compiler options, you can instead make a directory named like all.h.gch, and put each precompiled header in the directory, perhaps using -o. It doesn’t matter what you call the files in the directory; every precompiled header in the directory is considered. The first precompiled header encountered in the directory that is valid for this compilation is used; they’re searched in no particular order. 

So we just use that folder instead. Clang still won't care, since we're giving it the path explicitly, but gcc compile times are now back to what you'd expect. 

#### How to Verify:

Compile with `PRECOMPILED_HEADERS=false` and then with them enabled, compare timings. 

#### Copyright and Licensing

### DIFF
--- a/make/program
+++ b/make/program
@@ -28,8 +28,8 @@ $(PRECOMPILED_MODEL_HEADER) : $(STAN)src/stan/model/model_header.hpp
 	$(COMPILE.cpp) $< $(OUTPUT_OPTION)
 
 ifeq ($(CXX_TYPE),clang)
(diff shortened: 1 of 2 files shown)


rok-cesnovar (approved):
Very nice find! So the last versions had slower compile times with g++?

WardBrian:
Yes, and I guess it slipped under the radar

PR #1308, opened 2025-01-14
Change default sig_figs value to 8

### LINKED ISSUE #1307: Change the sig_figs default to 8
#### Summary

CmdStan stores with 6 digits, and Stan checks constraints with 1e-8. 

I propose we change the default CmdStan `sig_figs` to 8, to reduce constraint check failures due to current mismatch.

#### Description:

After `generate_quantities()` method was introduced, there have been several reports of problems with constraint checking. CmdStan stores the posterior draws with accuracy defined by `sig_figs` argument specifies the number of significant digits for all numerical values in the output files (documented at https://mc-stan.org/docs/cmdstan-guide/command_line_options.html#output-control.section, and mentioned also at https://mc-stan.org/docs/cmdstan-guide/stan_csv_apdx.html). The default for `sig_figs` is 6. When `generate_quantities()` method is called, the posterior draws are read from the file and constraint checking is made. The constraint checking is made with accuracy `CONSTRAINT_TOLERANCE`, which by default is 1e-8 as defined in [stan/math/prim/err/constraint_tolerance.hpp](https://github.com/stan-dev/math/blob/develop/stan/math/prim/err/constraint_tolerance.hpp). The discrepancy between stored 6 figures and constrain check with 1e-8, causes the constraint check to fail sometimes. Earlier this was a problem mostly with simplex as reported, e.g., in https://github.com/stan-dev/cmdstanr/issues/420. Now the zero-sum constraint seem to cause these errors more likely.

I propose we change the default CmdStan `sig_figs` to 8, to reduce constraint check failures due to current mismatch.

Pros:
- fewer users encounter confusing constraint check error
Cons:
- CmdStan csv files will be 33% bigger

Some users may also see slight changes in the results if they re-run some inference with new CmdStan version, but then bit-by-bit replicability from version to version has not been guaranteed before either.

Eventually binary output files would be a better solution, but that is a much bigger change.

WardBrian:
#### Submisison Checklist

#### Summary:

Closes #1307 

#### Intended Effect:

Increase the number of outputs saved by default

#### How to Verify:

Run a model without changing the sig_figs argument

#### Side Effects:

Additional storage may be required

#### Documentation:

TBD

#### Copyright and Licensing

Simons Foundation


mitzimorris:
also need to change the CmdStan docs

WardBrian:
> also need to change the CmdStan docs

Yes, will change several places if this gets merged

mitzimorris:
whatever you and Bob think is best.
cheers,
Mitzi

WardBrian:
Soliciting more opinions @bob-carpenter @SteveBronder @avehtari

avehtari:
Can you try with the example in https://github.com/stan-dev/cmdstanr/issues/420 (it's an issue in CmdStanR repo, but can be tested with any approach calling standalone generated quantities)?

WardBrian:
@avehtari for the specific example you link, it does appear to be resolved by this PR (passed over 100 times in a row)

src/cmdstan/arguments/arg_output_sig_figs.hpp
```diff
@@ -12,15 +12,15 @@ class arg_output_sig_figs : public int_argument {
     _description
         = "The number of significant figures used for the output CSV files.";
     _validity
-        = "0 <= integer <= 18 or -1 to use the default number of significant "
-          "figures";
-    _default = "-1";
-    _default_value = -1;
+        = "0 <= integer <= 18, or -1 to use the system library's default "
+          "number of significant figures (usually 6).";
```
bob-carpenter: shouldn't this `6` be an `8`?

src/cmdstan/arguments/arg_output_sig_figs.hpp
```diff
@@ -12,15 +12,15 @@ class arg_output_sig_figs : public int_argument {
     _description
         = "The number of significant figures used for the output CSV files.";
     _validity
-        = "0 <= integer <= 18 or -1 to use the default number of significant "
-          "figures";
-    _default = "-1";
-    _default_value = -1;
+        = "0 <= integer <= 18, or -1 to use the system library's default "
```
bob-carpenter: How about "an integer number of significant figures between 0 and 18 inclusive, or -1 to use the default (8)"

bob-carpenter (changes requested):
I think there's a slipped default, so I took the liberty of suggesting a wording change while you're at it.

src/cmdstan/arguments/arg_output_sig_figs.hpp
WardBrian: No, -1 is preserved to do the same thing as before, which is to just not specify any target precision when the stream is created, leaving it with whatever your OS defaults to.  This is why it’s the “system library’s default”, not ours — 6 comes from glibc and friends. 

If we want, we can make -1 synonymous with our default (8) instead

WardBrian: (Though I should note, if you want to use Stan’s default, a reasonable option is just not pass this argument at all…)

bob-carpenter: I don't think this is a huge backward compatiblity concern and it would certainly be simple to have just one default.  In which case, I don't think we'd need a behavior for -1, which would simplify the doc.

Having said that, I think it's OK for "unspecified" to be the Stan default and -1 the system default as long as it's a bit more clearly stated in the doc that the system default is the underlying system default, not the Stan default.

WardBrian: I would prefer to just not allow -1 at all, but removing a valid value would definitely break at least somebody’s code

bob-carpenter: Good point---I missed that was the old default!

WardBrian: I've updated the wording. Here's what the full help text reads as now:

```
    sig_figs=<int>
      The number of significant figures used for the output CSV files.
      Valid values: 0 <= integer <= 18, or -1 to use the current operating system's default number of significant figures (This will vary, but on many systems it is 6).
      Defaults to 8
```

bob-carpenter (approved):
I still find two things about this confusing:

* the use of "integer"---is that in other ones of these?
* there's no mention that the platform default is not the Stan default

But if you want to go with it like this, I'm just going to approve because it's not worth a lot of back and forth.

WardBrian:
> the use of "integer"---is that in other ones of these?

It doesn't look like it, I've updated it to be closer to the others.


```
    sig_figs=<int>
      The number of significant figures used for the output CSV files. 
      Stan recommends at least 8 (the default) if you will be using the output as part 
      of a future input (e.g., for standalone generated quantities or as an initialization). 
      If -1 is supplied, Stan will use the current operating system's default precision 
      (This typically be smaller than Stan's default; on many systems it is 6 digits).
      Valid values: 0 <= sig_figs <= 18 || sig_figs == -1
      Defaults to 8
```

PR #1313, opened 2025-02-27
Remove unused and outdated github actions script

WardBrian:
This will close #1295 as well


andrjohns (approved):
Deleting code has got to be one of my favourite things

bob-carpenter:
> Deleting code has got to be one of my favourite things.

Me, too!  Of all my PRs since version 1, I'm most happy with the autodiff testing framework, that massively reduced test size while extending coverage.  These are PRs only a dev can love, since they're not user facing.  So thanks for the housekeeping, @WardBrian.

PR #1314, opened 2025-03-10
Clean up test makefile dependencies

WardBrian:
#### Submisison Checklist

#### Summary:

This cleans up some bad habits in how our makefile for the tests was written, to try to fix the odd dependency race condition we are seeing in the upstream tests for https://github.com/stan-dev/math/pull/3159.

#### Intended Effect:

Make tests build more reliably. 

#### Copyright and Licensing

### DIFF
--- a/make/tests
+++ b/make/tests
@@ -41,7 +41,7 @@ ifneq ($(filter test/%,$(MAKECMDGOALS)),)
 
 include src/cmdstan/main.d
 
-TEST_SRCS = $(wildcard src/test/*.cpp)
+TEST_SRCS = $(wildcard src/test/interface/*.cpp)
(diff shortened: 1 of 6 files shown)


andrjohns (approved):
Small question - why were the header changes needed?

But all LGTM either way!

WardBrian:
> why were the header changes needed?

Those tests only need the constants defined in return_code, but were including all of the cmdstan code, including the forward declaration of the functions in a generated model, so they needed to include a generated model... this was all ONLY to make the linker happy, not because that code was actually needed for the test!

Not having them include generated code makes the make logic easier and avoids some unnecessary re-builds in some circumstances

andrjohns:
Ahhh that makes sense, yeah that's a nice cleanup!

PR #1360, opened 2026-09-23
Apply subst to USER_HEADER

ndabas:
#### Submisison Checklist

#### Summary:

Currently, `make/program` applies `$(subst \,/,...)` to every path it hands to a tool, with one exception: `USER_HEADER` on line 47.

On Windows this matters because Stan Math sets `SHELL` when a Unix shell is available, which it always is with RTools on `PATH`.

Assigning `SHELL` in a `makefile` takes GNU Make off its direct-exec fast path, so recipes are run through `sh`, and `sh` strips the backslashes from the unquoted `-include` argument. The compiler is then handed a path with no separators at all and fails to find the header:

```
<built-in>:1:10: fatal error: 'C:UsersndabaDocumentsGitHubcmdstanpytestdatareturn_one.hpp' file not found
    1 | #include "C:UsersndabaDocumentsGitHubcmdstanpytestdatareturn_one.hpp"
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
make: *** [make/program:83: .../external.o] Error 1
```

This is a shell-quoting problem, not a compiler problem: replacing the compiler with an argv-dumping stub in the failing build shows the separators are already gone by the time the compiler is invoked, and both clang and GCC accept the backslash form when it reaches them intact.

This came up in stan-dev/cmdstanpy#864, where `cmdstanpy` currently works around it by converting `user_header` to a POSIX path before passing it to make.

#### Intended Effect:

`-include` receives a forward-slash path, matching how every other path in this file is passed to a tool. No behaviour change on Unix, or on Windows where `USER_HEADER` is already given with forward slashes.

#### How to Verify:

```sh
make STANCFLAGS=--allow-undefined USER_HEADER=C:\path\to\user_header.hpp C:/path/to/model.exe
```

Before: `fatal error: '...' file not found`, with the separators stripped from the path. After: compiles and links.

Verified on Windows ARM64 with RTools 4.5 aarch64 (clang 19.1.7, GNU Make 4.4.1) against CmdStan 2.40.0, using `external.stan` / `return_one.hpp` from cmdstanpy's test data: the build fails before the change and succeeds after it, with the
emitted flag going from `-include C:\Users\...\return_one.hpp` to `-include C:/Users/.../return_one.hpp`.

#### Side Effects:

None expected. `$(subst ...)` is a literal separator swap, so it is a no-op for paths that contain no backslashes.

The remaining bare uses of `$(USER_HEADER)` in this file are deliberate: the prerequisite on line 80 and the missing-header diagnostic on lines 51-57, which should echo the path as the user supplied it.

#### Documentation:

No documentation changes needed.

#### Copyright and Licensing

Nikhil Dabas


WardBrian (approved):
Thanks!

